PR #12, opened 2019-09-16
remove extra rule for git.apache.org

pb82:
Remove the extra rule that was required for go < 1.12

Verification steps:

1. Make sure that `make code/compile` and `make test/unit` still work with this branch


abstractj:
@pb82 could you please run `go mod tidy` and resubmit the changes?

pb82:
@abstractj Hmm, i did that. Its still failing the analyze step

abstractj:
@pb82 I'm afraid to say that we may need this. I'm getting a timeout to download thrift, just running `go build` locally.

slaskawi:
@abstractj @pb82 @davidffrench I know we've been talking about this one, but maybe we should commit that `vendor` directory?

davidffrench:
@abstractj It looks like this rule needs to be removed to get the commands to work, this is happening on Go 1.13 for all 3 of us. 

@slaskawi Could be right, I think we need to start committing in the vendor directory to avoid the issue reported by @abstractj .

abstractj:
> @abstractj It looks like this rule needs to be removed to get the commands to work, this is happening on Go 1.13 for all 3 of us.
> 
> @slaskawi Could be right, I think we need to start committing in the vendor directory to avoid the issue reported by @abstractj .
> 
> I don't think this will fix the issue with `GolangCI`, @abstractj is there any way for us to find out why this is failing?

For more that I love to work on the latest greatest releases. Go 1.13 was released in September 3rd, packages are not available yet to most of the distributions and the minimum version supported by operator-sdk is 1.12. 

Are we going to target Go 1.13? I afraid that we may spend more time struggling with the issues of a new release of Go, instead of actually thinking about what we want to achieve with the Keycloak operator.

My suggestion is to hold this upgrade, until we are 100% that Go 1.13 is properly distributed in most of the distros. 

If you guys really think we should go with 1.13 and have a consensus about this, it's ok. But I find it a bit risky.

slaskawi:
@abstractj @pb82 Here's another try to fix the dependency problem: https://github.com/keycloak/keycloak-operator/pull/15

PR #13, opened 2019-09-16
Intly 3360

pb82:
## JIRA ID

INTLY-3360

## Additional Information
Adds the spec for the Keycloak CRD


deploy/crds/Keycloak.yaml
```diff
@@ -29,6 +29,28 @@ spec:
...
+              type: string
+              required: false
+            extensions:
+              description: 'A list of extensions to be loaded by Keycloak on startup'
+              type: array
+              items:
+                type: object
+                properties:
+                  name:
+                    description: 'Extension name'
```
pb82: There is currently no repository for keycloak extensions. I wonder, should we instead ask for a download URL?

Or use the old approach where we bundle the extensions with the init container?

deploy/crds/Keycloak.yaml
```diff
@@ -29,6 +29,28 @@ spec:
           type: object
         spec:
           type: object
+          properties:
+            externalDatabaseSecret:
+              description: 'Opotional. A Secret containing the connection string and credentials to an external database. If not provided a Postgres
```
davidffrench:
```suggestion
              description: 'Optional. A Secret containing the connection string and credentials to an external database. If not provided a Postgres
```

davidffrench: Should this description also list the properties required in the secret or a link to where they can get that information?

deploy/crds/Keycloak.yaml
davidffrench: As discussed on the call, lets use the old approach for now. So using the old approach, what should be changed in the spec? To list the initcontainer image name and location maybe?

pb82: For the old approach we only need the names of the extensions. The init container image and tag are hardcoded in the operator unless we want to make it configurable.

deploy/crds/Keycloak.yaml
pb82: Yep, good point.

deploy/crds/Keycloak.yaml
davidffrench: ok, perfect. And the names of the extensions are used to determine what is added through the init container?

deploy/crds/Keycloak.yaml
```diff
@@ -29,6 +29,28 @@ spec:
...
+                instance will be created in the same namespace'
+              type: string
+            adminCredentialSecret:
```
slaskawi: The main issue I have with those references to secrets is how to tell what is the structure of a secret? What fields it should have? Is there any pattern to document things like that?

deploy/crds/Keycloak.yaml
slaskawi: > There is currently no repository for keycloak extensions. I wonder, should we instead ask for a download URL?

Currently there is no mechanism in Keycloak (nor RHSSO) images to download and install extensions on the fly. I'm not sure if there should be any (as this is a potential attack vector). So I think we should provide our own way of installing them. I'm OK with the old approach. Just wanted to check, what do you think about using Maven to download them?

deploy/crds/Keycloak.yaml
```diff
@@ -29,6 +29,28 @@ spec:
...
+              items:
+                type: object
+                properties:
+                  name:
+                    description: 'Extension name'
+                    type: string
+                  version:
+                    description: 'Extension version'
+                    type: string
         status:
```
slaskawi: I think we should also have a number of instances (since we'll be running HA configuration by default).

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -10,9 +10,14 @@ import (
...
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+	ExternalDatabaseSecret string              `json:"externalDatabaseSecret,omitempty"`
+	AdminCredentialSecret  string              `json:"adminCredentialSecret,omitempty"`
+	Extensions             []KeycloakExtension `json:"extensions,omitempty"`
+}
+
+type KeycloakExtension struct {
+	Name    string `json:"name"`
```
slaskawi: I think there should be either full URL or Maven coordinates here (`G:A:V`).

deploy/crds/Keycloak.yaml
pb82: It should be in two places:
1. The CRD (openapi validation section) as @davidffrench pointed out
1. The documentation of the operator

pkg/apis/keycloak/v1alpha1/keycloak_types.go
pb82: Ok, i'll change the type of `extensions` to a list of strings. We can then probe the value in the operator to figure out if its either a valid URL or Maven Group:Artifact:Version.

deploy/crds/Keycloak.yaml
pb82: Done.

pkg/apis/keycloak/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -36,6 +36,22 @@ func (in *Keycloak) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *KeycloakExtension) DeepCopyInto(out *KeycloakExtension) {
```
golangcibot: undeclared name: `KeycloakExtension` (from `typecheck`)

PR #17, opened 2019-09-17
INTLY-3222 Use vendor by default (Go 1.12)

slaskawi:
## JIRA ID

INTLY-3222

## Additional Information

It seems Go 1.12 requires adding `-mod=vendor` flag.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ code/run:
 
 .PHONY: code/compile
 code/compile:
-	@GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=${CGO_ENABLED} go build -o=$(COMPILE_TARGET) ./cmd/manager
+	@GOOS=${GOOS} GOARCH=${GOARCH} CGO_ENABLED=${CGO_ENABLED} go build -o=$(COMPILE_TARGET) -mod=vendor ./cmd/manager
 
 .PHONY: code/gen
 code/gen:
@@ -63,7 +63,7 @@ image/build/push: image/build image/push
 .PHONY: test/unit
 test/unit:
 	@echo Running tests:
-	go test -v -race -cover ./pkg/...
+	go test -v -race -cover -mod=vendor ./pkg/...
 
 .PHONY: code/lint
 code/lint:


abstractj:
@slaskawi it's not because if Go, it's because we added the `vendor` folder. We don't have it on Gatekeeper https://github.com/keycloak/keycloak-gatekeeper/blob/master/Makefile#L137

davidffrench:
👀

davidffrench:
Sorry, should have commented on this yesterday. I have ran the make commands locally and they are working correctly with `-mod=vendor`, they also seem to be working fine without it. Is there something that I am missing here?

Does this need to be added to the `code/fix` and `code/check` commands as well?

slaskawi:
@davidffrench I think you might be using Go `1.13`, is that correct? If so, the `vendor` directory will be picked up automatically. With Go `1.12`, you need the additional switch. 

Please compare this two Travis runs:
- Our master build, please notice that it downloads all dependencies: https://travis-ci.org/keycloak/keycloak-operator/builds/586045508
- This PR, please notice, it doesn't download anything: https://travis-ci.org/keycloak/keycloak-operator/builds/585982363?utm_source=github_status&utm_medium=notification

davidffrench:
@slaskawi I have already switched to Go`1.12`. Using `go version go1.12.9 darwin/amd64`. It is strange that it is using the vendor folder for me locally without having to specify this.

After doing some reading, you are correct, it does say it is necessary. https://github.com/golang/go/wiki/Modules
> If vendoring is expected to be used, check that the -mod=vendor flag is being passed to go build or similar, or that GOFLAGS=-mod=vendor is set.
> Modules by default ignore the vendor directory unless you ask the go tool to use vendor.

Approving this PR.

slaskawi:
>  have already switched to Go1.12. Using go version go1.12.9 darwin/amd64. It is strange that it is using the vendor folder for me locally without having to specify this.

@davidffrench The only thing that comes into my mind is that maybe your IDE somehow uses different Go version (maybe embedded or something). You can also try to clean all the caches: `go clean -cache -modcache -i -r`

PR #21, opened 2019-09-20
INTLY-3543 provide reconciler implementation and unit test

pb82:
## JIRA ID

INTLY-3543

Provides an implementation for the reconciler based on the discussed approach:

In every reconciliation we determine the current cluster state.
The current cluster state combined with the Keycloak CR is used to generate the desired cluster state.
The desired cluster state is a list of actions that need to be run in order to get the cluster from the current state into the desired state.

Those actions are then passed to an `ActionRunner` which will actually perform them against the cluster.

There are two implementations of `ActionRunner`: `ClusterActionRunner` for performing real actions and `MockActionRunner` for Unit tests (which only count the number of updates, creates, etc.)

A sample unit test is provided.

PR #22, opened 2019-09-23
Intly 3543 reconciliation loop

slaskawi:
## JIRA ID

INTLY-3543

## Additional Information

(copy-pasted from https://github.com/keycloak/keycloak-operator/pull/21)

Provides an implementation for the reconciler based on the discussed approach:

In every reconciliation we determine the current cluster state.
The current cluster state combined with the Keycloak CR is used to generate the desired cluster state.
The desired cluster state is a list of actions that need to be run in order to get the cluster from the current state into the desired state.

Those actions are then passed to an `ActionRunner` which will actually perform them against the cluster.

There are two implementations of `ActionRunner`: `ClusterActionRunner` for performing real actions and `MockActionRunner` for Unit tests (which only count the number of updates, creates, etc.)

A sample unit test is provided.

## Checklist:
- [ ] Verified by team member
- [x] Automated Tests

## Additional Notes

In addition to @pb82 changes, I made the reconcilation loop more explicit. I believe this approach is much better for navigating over the code. 

Also, I simplified the tests. Now we don't need any mocks (and this is exactly what I wanted).


pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -0,0 +1,42 @@
...
+	kc "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	"github.com/keycloak/keycloak-operator/pkg/common"
+	"github.com/keycloak/keycloak-operator/pkg/model/keycloak"
+)
+
+type Reconciler interface {
+	Reconcile(clusterState *common.ClusterState, cr *kc.Keycloak) (common.DesiredClusterState, error)
+}
+
+type KeycloakReconciler struct {
```
golangcibot: type name will be used as keycloak.KeycloakReconciler by other packages, and that stutters; consider calling this Reconciler (from `golint`)

slaskawi: This one is definitely false-positive. We will probably also have `RHSSOReconciler` in the same file (or somewhere nearby).

slaskawi: Alternatively, I can move it to `controller` (one level up). But then, we'll have one fat package with all stuff in it. I'm not sure if I like it.

davidffrench: I would much prefer to leave the name as `KeycloakReconciler`, happy to leave this lint on as it would be useful in other places and possibly bypass it with a comment if that is possible with GolangCI?

slaskawi: I tried to look it up but unfortunately I haven't found anything. Maybe @abstractj knows about anything like that.

slaskawi: ooh wait... just found this: `// nolint`

slaskawi: Works like a charm!

pkg/common/cluster_state.go
```diff
@@ -0,0 +1,43 @@
...
+// Keycloak service
+func (i *ClusterState) readKeycloakServiceCurrentState(cr *kc.Keycloak, controllerClient client.Client) {
+	keycloakService := keycloak.Service(cr)
+
+	selector := client.ObjectKey{
+		Name:      keycloakService.Name,
+		Namespace: keycloakService.Namespace,
+	}
+	err := controllerClient.Get(context.TODO(), selector, keycloakService)
+	if err != nil {
```
davidffrench: The way I am reading this error handling is if there is no error returned from the `Get`, the current cluster state for the keycloakservice will be set to `nil`, if there is an error, it set the cluster state to the keycloakService object. 

This should be reversed and also handle the difference between an error and when the resource doesn't exist. 

I haven't been able to find much from the docs on what is returned when the resource doesn't exist. https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#DelegatingReader.Get

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	"github.com/keycloak/keycloak-operator/pkg/common"
+	"github.com/keycloak/keycloak-operator/pkg/model/keycloak"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestKeycloakReconciler_Test_Creating_Example_Service(t *testing.T) {
```
davidffrench: Would be good to add an update test here as well.

slaskawi:
> Having to specify both a Run function and a struct for all none generic actions feels like additional code that could be simplified. We can talk about this on the stand up.

@davidffrench As we agreed on the meeting, we'll just remove `ServiceUpdateAction` and implement a more sophisticated actions once we need them.

> We should start getting tests in for the cluster_actions using https://godoc.org/k8s.io/client-go/kubernetes/fake soon.

@davidffrench I looked into this library and it looks nice. However, we can easily test K8s objects in unit tests and create/update actions in end-to-end tests. It's also important to note, that we'll need 2 types e2e tests - for vanilla Kubernetes and for OpenShift. So how about leaving this for now, and once we have something that works (e.g. database deployment), we'll try to test it out in end-to-end test and see how it goes?

> We need to add some work to get the status of the CR updated as well. I think this could be a separate piece of work to begin with though, using https://github.com/redhat-cop/operator-utils

@davidffrench During the meeting we agreed to implement it in a separate PR.

pkg/controller/keycloak/keycloak_reconciler_test.go
slaskawi: Done

pkg/common/cluster_state.go
slaskawi: I looked through the code and it indeed seems fine :fearful: 

The `Get` method tries to obtain an objects from a cache (see [here](https://github.com/keycloak/keycloak-operator/blob/4dfd3d9d1b8e6011c0c50ab7b4b08f4804046677/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go#L53)). If the resources, doesn't exist - it returns an error. 

So in short, if there's an error (note the `!=` in the code), the object doesn't exist. Therefore, we set the `i.KeycloakService = nil`.

pkg/common/cluster_actions.go
```diff
@@ -0,0 +1,74 @@
...
+	Msg string
+}
+
+func (i GenericCreateAction) Run(runner ActionRunner) (string, error) {
+	return i.Msg, runner.Create(i.Ref)
+}
+
+func (i GenericUpdateAction) Run(runner ActionRunner) (string, error) {
+	return i.Msg, runner.Update(i.Ref)
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
}
```

pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if clusterState.KeycloakService == nil {
+		return common.GenericCreateAction{
+			Ref: service,
+			Msg: "create keycloak service",
+		}
+	}
+
+	return common.GenericUpdateAction{
+		Ref:             service,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

pkg/common/cluster_state.go
davidffrench: Thanks @slaskawi . You are correct, I was reading this wrong. So in the case here where an error is returned, we need to handle legitimate errors as well. We should differentiate Not Found errors which will set it to nil and all other errors should return an error code. This can be done by checking the error type is `NewNotFound`. https://github.com/keycloak/keycloak-operator/blob/4dfd3d9d1b8e6011c0c50ab7b4b08f4804046677/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/cache_reader.go#L61

pb82: There is a helper method for that in the `errors` package (IsNotFoundError)

abstractj:
@slaskawi I'm still reading through the code but at first glance I'd suggest to change the commit message to "[INTLY-3543] Create desired and current state objects and helper methods". I got a bit lost on trying to match the Jira with the description and also the goal of this PR.

pkg/common/cluster_state.go
slaskawi: Implemented.

pkg/common/cluster_state.go
```diff
@@ -0,0 +1,50 @@
...
+
+func NewClusterState() *ClusterState {
+	return &ClusterState{
+		KeycloakService: nil,
+	}
+}
+
+func (i *ClusterState) Read(cr *kc.Keycloak, controllerClient client.Client) error {
+	i.readKeycloakServiceCurrentState(cr, controllerClient)
+}
```
golangcibot: missing return (from `typecheck`)

slaskawi:
> I'm still reading through the code but at first glance I'd suggest to change the commit message to "[INTLY-3543] Create desired and current state objects and helper methods". I got a bit lost on trying to match the Jira with the description and also the goal of this PR.

I believe you're referring to https://github.com/keycloak/keycloak-operator/pull/22/commits/53715041c289d440cb36c11f63de95c754ce571a ? That's @pb82 's commit and I would override his name if I changed it. If it's fine with you, I would prefer to leave it as is.

pkg/common/cluster_state.go
slaskawi: Fixed.

slaskawi:
@davidffrench I also added testing Actions to this PR.

pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if clusterState.KeycloakService == nil {
+		return common.GenericCreateAction{
+			Ref: service,
+			Msg: "create keycloak service",
+		}
+	}
+
+	return common.GenericUpdateAction{
+		Ref: service,
```
davidffrench: The service would need to have the clusterIP and resourceversion added to it.

slaskawi: Could you please explain, why we need both of them (`resourceVersion` and `clusterIP`)? As far as I know, the API server automatically sets the `resourceVersion` and using a Cluster IP as a Service Type is the default one (other types are `NodePort`, `LoadBalancer` and headless - `None`).

davidffrench: Unfortunately I don't have the answer, this was brought up by Peter on his PR that for some reason Service resource updates require the ClusterIP and ResourceVersion to be present. 

I have not done any research into this myself. It should be very easy to test though, I will run your operator locally and see if there is any errors.

slaskawi: Ok, great! If you encounter any problems - please let me know.

davidffrench: Just tested this locally and am getting the following error on the second reconcile loop `{"level":"error","ts":1569320357.343873,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"keycloak-controller","request":"keycloak/example-keycloak","error":"Service \"keycloak\" is invalid: [metadata.resourceVersion: Invalid value: \"\": must be specified for an update, spec.clusterIP: Invalid value: \"\": field is immutable]","stacktrace":"github.com/keycloak/keycloak-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/keycloak/keycloak-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:217\ngithub.com/keycloak/keycloak-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158\ngithub.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/dffrench/go/src/github.com/keycloak/keycloak-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}`

slaskawi: Fixed. Thanks!

davidffrench: Perfect, tested again and this is working correctly.

PR #25, opened 2019-09-24
[KEYCLOAK-11498] Add Coveralls to CI

abstractj:
## Additional Information
* Included tasks into the Makefile to centralize everything in a single place
* `make/coverage` may overlap with `make/unit` so we might want to consider to merge then in a single command

## Verification Steps
* See: https://coveralls.io/github/abstractj/keycloak-operator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: go
-
 matrix:
   include:
   - go: 1.12.x
     env: GO111MODULE=on
-    
 script:
 - make test/unit
+- make code/goveralls
--- a/Makefile
+++ b/Makefile
@@ -96,3 +96,18 @@ code/lint:
 setup/travis:
 	@echo Installing Operator SDK
 	@curl -Lo operator-sdk ${OPERATOR_SDK_DOWNLOAD_URL} && chmod +x operator-sdk && sudo mv operator-sdk /usr/local/bin/
+
+.PHONY: code/coverage
+code/coverage:
+	@echo "--> Running go coverage"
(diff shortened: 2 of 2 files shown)


davidffrench:
Looks good @abstractj . I would be fine with the commands being combined and the `make test/unit` command generating a coverage report all the time.

Can we also get coveralls to comment on the PRs. there should be an option in settings for this, I am not sure if it is turned on by default.

abstractj:
> Looks good @abstractj . I would be fine with the commands being combined and the `make test/unit` command generating a coverage report all the time.

I will update this PR and move my changes to `test/unit`

> 
> Can we also get coveralls to comment on the PRs. there should be an option in settings for this, I am not sure if it is turned on by default.

It's up to you guys to decide on this, but I have several cons about doing this and also code coverage:

1. Code coverage is a great information to have, but at the same time, easy to cheat. You can achieve 99.9999% of test coverage and have silly tests only because you want to achieve the maximum of coverage.
2. A bot commenting on how much coverage a change introduced or decreased, may generate a lot of noise. IMO the only relevant bot should be the GolangCI

But that's just my opinion, if you guys decide that's good for the project. At the moment, I just enabled it and you guys let me know if should be disabled.

slaskawi:
> A bot commenting on how much coverage a change introduced or decreased, may generate a lot of noise. IMO the only relevant bot should be the GolangCI

@abstractj hmmmm this actually sounds very interesting to me. Personally, I would like to have it but with a caveat that decreasing coverage is **OK**. I don't want any pressure about the coverage in initial development stage.

PR #30, opened 2019-09-25
Minor fix for Makefile

abstractj:
The command `make code/compile` will fail if project variable is not provided. In order to fix this, we just need to add the project name on top of the Makefile

I didn't create any Jiras because it's just a minor fix


slaskawi:
@abstractj Could you please check why Golang CI failed?

abstractj:
@slaskawi sure np.

PR #36, opened 2019-09-30
[KEYCLOAK-11536] Keycloak Operator Dockerfile fails to build

abstractj:
## Verification Steps
* Initial tests done with quay.io here: https://quay.io/repository/abstractj/keycloak-operator/build/5f8fccb5-9640-46b4-ac44-6f858b9d55a6

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,14 +1,21 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
 
+ENV GOCACHE=/go/path/.cache \
+    GOPATH=/go/path \
+    GOROOT=/usr/local/go
+ENV PATH=$PATH:$GOROOT/bin:$GOPATH/bin
+
 ENV OPERATOR=/usr/local/bin/keycloak-operator \
     USER_UID=1001 \
     USER_NAME=keycloak-operator
 
-# install operator binary
-COPY build/_output/bin/keycloak-operator ${OPERATOR}
+COPY tools /opt/jboss/tools
+RUN /opt/jboss/tools/build-operator.sh
 
-COPY build/bin /usr/local/bin
+COPY bin /usr/local/bin
 RUN  /usr/local/bin/user_setup
+
+RUN rm -rf /usr/local/go && rm -rf /go && rm -rf /opt/jboss/tools 
 
 ENTRYPOINT ["/usr/local/bin/entrypoint"]
 
--- a/build/tools/build-operator.sh
+++ b/build/tools/build-operator.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+# Default Keycloak repository and destination path
+# In the future we may want to provide it as arguments to the build
+
+GIT_REPO="https://github.com/keycloak/keycloak-operator.git"
+KEYCLOAK_PATH="/go/src/github.com/keycloak"
+
+mkdir -p $KEYCLOAK_PATH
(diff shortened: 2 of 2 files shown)


abstractj:
> hmmm this approach is a bit surprising...
> 
> Could you please explain, why don't we just copy Operator's binary into a simple image (might be even a unikernel), like [for example Prometheus Operator does](https://github.com/coreos/prometheus-operator/blob/master/Dockerfile)?

@slaskawi in order to copy the binary into a simple image, we need a release of this binary. Something which we don't have at the moment. As soon as we have a release of the binary, yes, we can split it into release image and dev image.

Another alternative is to use operator-sdk to build the Docker image. That would require to:
1. Build the Docker image using operator-sdk on Travis after all the tests pass
2. Provide access tokens from quay.io to Travis in order to push this image

That's the standard for most the Operators out there, but @stianst would like to keep exactly the same process we have today for other images. That means our builds are triggered by a GH push and quay.io does all the work. For more context, please see the whole discussion here: https://issues.jboss.org/browse/KEYCLOAK-11354.

slaskawi:
@abstractj @stianst Ok, that sounds sensible to me. We can try it out. 

However, I have two major comments - we probably should install Golang from an RPM. Otherwise it will be pretty hard to download a binary during productization. Also, I think we should clean up everything we installed for compiling the application. This will make us more immutable for CVEs (and that's another argument for installing Golang from an RPM).

abstractj:
> @abstractj @stianst Ok, that sounds sensible to me. We can try it out.
> 
> However, I have two major comments - we probably should install Golang from an RPM. Otherwise it will be pretty hard to download a binary during productization. Also, I think we should clean up everything we installed for compiling the application. This will make us more immutable for CVEs (and that's another argument for installing Golang from an RPM).

We are using ubi-minimal image, which means we have a very limited access to some RPMs, so `microdnf` won't find golang package. But if you know how to do that, I will be more than happy do add.

The clean up is already being performed, please check the build script. The script do the clean ups and also the removal of installation files for Go.

In summary we gonna end up with two kinds of images:

1. Imagine with the binary file - this will pretty much download the binary from GitHub or whateve place we choose and generate a Docker image. Same approach as we do for Gatekeeper https://github.com/keycloak/keycloak-containers/blob/master/gatekeeper/Dockerfile
2. The approach used in this PR. Build the binaries from scratch and use the latest greatest changes based on `master`

slaskawi:
> We are using ubi-minimal image, which means we have a very limited access to some RPMs, so microdnf won't find golang package. But if you know how to do that, I will be more than happy do add.

@abstractj I see. Indeed, the Golang is not present in RHEL image.

> The clean up is already being performed, please check the build script. The script do the clean ups and also the removal of installation files for Go.

@abstractj Sorry I wasn't precise in my previous comment. Once we compile the Operator inside the image, we could wipe the Go binaries as well as cloned sources out. They are no longer needed.

> In summary we gonna end up with two kinds of images:

@abstractj Ok that sounds fair. 

I also did some research on this topic and it seems we could leverage another approach - Builder Images (see this [link](https://fedoramagazine.org/building-smaller-container-images/)). This way we would create a builder image with Golang (and everything else needed for creating a binary) and then just copy the result into a fresh image. 

The end result would be the same - a Dockerhub (or Quay) building will build the Keycloak Operator binary but once it's done - it will copy it (without Golang dependencies or cloned repo) into a fresh image. This will result in a smaller and cleaner final image. I think it might be worth the effort. //cc @davidffrench @pb82 @stianst 

Once note - we'd need to check if this approach is OK with the productization process.

PR #41, opened 2019-10-04
INTLY-3225 Use OpenShift certificate if possible

slaskawi:
## JIRA ID

[INTLY-3225](https://issues.jboss.org/browse/INTLY-3225)

## Additional Information
This PR configures Keycloak to use OpenShift generated certificates for HTTPS. This way, other services may trust Keycloak by using `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/model/keycloak_deployment.go
+++ b/pkg/model/keycloak_deployment.go
@@ -36,6 +36,17 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
 					},
 				},
 				Spec: v1.PodSpec{
+					Volumes: []v1.Volume{
+						{
+							Name: ServingCertSecretName,
+							VolumeSource: v1.VolumeSource{
+								Secret: &v1.SecretVolumeSource{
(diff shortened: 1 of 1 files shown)


pb82:
@slaskawi The presence of the Serving Cert Secret seems to be OpenShift speficic (see https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_service.go#L21). I wonder if it makes sense to enable this only through a flag?

slaskawi:
> The presence of the Serving Cert Secret seems to be OpenShift speficic (see https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_service.go#L21). I wonder if it makes sense to enable this only through a flag?

@pb82 This is intentional. This is where mounting a secret with "Optional" flag set to true kicks in. If the secret is not there, Keycloak will generate a self-signed certificate upon the first HTTPS request.

PR #43, opened 2019-10-08
INTLY-3765 RHSSO deployment

slaskawi:
## JIRA ID

[INTLY-3765](https://issues.jboss.org/browse/INTLY-3765)

## Additional Information

This Pull Request adds support for RHSSO by using a program argument - `--profile RHSSO`.

## Checklist:
- [ ] Verified by team member
- [ ] Comments where necessary
- [ ] Automated Tests
- [ ] Documentation changes if necessary

PR #44, opened 2019-10-08
feat: Add logic to update Keycloak status

davidffrench:
## JIRA ID
https://issues.jboss.org/browse/INTLY-3642

## Additional Information
Adding logic to update the Keycloak status. Updates on success and error
- Updates list of secondary resources
- Updates ready boolean
- Updates phase
- Updates phase and message when an error occurs

## Verification Steps
1. Run operator from this branch
2. Create a Keycloak CR
3. Verify the status in the CR is updated correctly

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -42,10 +42,10 @@ type KeycloakStatus struct {
 type StatusPhase string
 
 var (
-	NoPhase          StatusPhase
-	PhaseReconciling StatusPhase = "reconciling"
-	PhaseFailing     StatusPhase = "failing"
-	PhaseInProgress  StatusPhase = "inprogress"
+	NoPhase           StatusPhase
+	PhaseReconciling  StatusPhase = "reconciling"
+	PhaseFailing      StatusPhase = "failing"
+	PhaseInitialising StatusPhase = "initialising"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -72,4 +72,20 @@ type KeycloakList struct {
 
 func init() {
 	SchemeBuilder.Register(&Keycloak{}, &KeycloakList{})
+}
+
+func (k *Keycloak) UpdateStatusSecondaryResources(instance *Keycloak, kind string, resourceName string) {
+	// If the map is nil, instansiate it
+	if instance.Status.SecondaryResources == nil {
+		instance.Status.SecondaryResources = make(map[string][]string)
+	}
+
+	// return if the resource name already exists in the slice
+	for _, ele := range instance.Status.SecondaryResources[kind] {
+		if ele == resourceName {
+			return
+		}
+	}
+	// add the resource name to the list of secondary resources in the status
+	instance.Status.SecondaryResources[kind] = append(instance.Status.SecondaryResources[kind], resourceName)
 }
--- a/pkg/common/cluster_actions.go
+++ b/pkg/common/cluster_actions.go
@@ -3,6 +3,8 @@ package common
 import (
 	"context"
 	"fmt"
+	"reflect"
+
 	"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -59,7 +61,14 @@ func (i *ClusterActionRunner) Create(obj runtime.Object) error {
 		return err
 	}
 
-	return i.client.Create(i.context, obj)
+	err = i.client.Create(i.context, obj)
+	if err != nil {
+		return err
+	}
+
+	i.cr.UpdateStatusSecondaryResources(i.cr, reflect.TypeOf(obj).String()[1:], obj.(v1.Object).GetName())
+
+	return nil
 }
 
 func (i *ClusterActionRunner) Update(obj runtime.Object) error {
--- a/pkg/common/cluster_state.go
+++ b/pkg/common/cluster_state.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"errors"
 
 	v13 "github.com/openshift/api/route/v1"
 	"k8s.io/api/extensions/v1beta1"
@@ -12,7 +13,7 @@ import (
 	"github.com/keycloak/keycloak-operator/pkg/model"
 	v12 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -26,6 +27,7 @@ const (
 	IngressKind               = "Ingress"
 	DeploymentKind            = "Deployment"
 	PersistentVolumeClaimKind = "PersistentVolumeClaim"
+	ConditionStatusSuccess    = "True"
 )
 
 // The desired cluster state is defined by a list of actions that have to be run to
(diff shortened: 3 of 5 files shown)


pkg/common/cluster_actions.go
```diff
@@ -59,7 +61,13 @@ func (i *ClusterActionRunner) Create(obj runtime.Object) error {
 		return err
 	}
 
-	return i.client.Create(i.context, obj)
+	err = i.client.Create(i.context, obj)
+	if err != nil {
+		return err
+	}
+
+	i.cr.UpdateStatusSecondaryResources(i.cr, reflect.TypeOf(obj).String(), obj.(v1.Object).GetName())
```
davidffrench: Does anyone have any suggestions on a better wat to get the kind other than `reflect.TypeOf(obj).String()`? This currently outputs "*v1.Service" for example with the `*`.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -45,7 +45,7 @@ var (
 	NoPhase          StatusPhase
```
davidffrench: The version is not updated in the status. Need to find a way to get this dynamically. Any suggestions?

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -45,7 +45,7 @@ var (
...
 	PhaseReconciling StatusPhase = "reconciling"
 	PhaseFailing     StatusPhase = "failing"
-	PhaseInProgress  StatusPhase = "inprogress"
+	PhaseInitalising StatusPhase = "initalising"
```
golangcibot: `initalising` is a misspelling of `initialising` (from `misspell`)
```suggestion

```

pkg/common/cluster_state.go
```diff
@@ -349,11 +350,77 @@ func (i *ClusterState) readKeycloakIngressCurrentState(context context.Context,
...
+	}
+
+	return keycloakDeploymentReady && postgresqlDeploymentReady && keycloakRouteReady, nil
+}
+
+func (i *ClusterState) isDeploymentReady(deployment *v12.Deployment) (bool, error) {
+	if deployment == nil {
+		return false, nil
+	}
+	// A deployment has an array of condidtions
```
golangcibot: `condidtions` is a misspelling of `conditions` (from `misspell`)
```suggestion

```

pkg/common/cluster_state.go
```diff
@@ -349,11 +350,77 @@ func (i *ClusterState) readKeycloakIngressCurrentState(context context.Context,
...
+	for _, condition := range deployment.Status.Conditions {
+		// One failure condition exists, if this exists, return the Reason
+		if condition.Type == v12.DeploymentReplicaFailure {
+			return false, errors.New(condition.Reason)
+			// A successful deployment will have the progressing condition type as true
+		} else if condition.Type == v12.DeploymentProgressing && condition.Status != "True" {
```
golangcibot: string `True` has 2 occurrences, make it a constant (from `goconst`)

pb82:
@davidffrench this looks good, the following was written into the status:

```
status:
  message: ''
  phase: reconciling
  ready: true
  secondaryResources:
    '*v1.Route':
      - keycloak
    '*v1.PersistentVolumeClaim':
      - keycloak-postgresql-claim
    '*v1alpha1.GrafanaDashboard':
      - keycloak
    '*v1.Service':
      - keycloak-postgresql
      - keycloak
      - keycloak-discovery
    '*v1.StatefulSet':
      - keycloak
    '*v1.Deployment':
      - keycloak-postgresql
    '*v1.ServiceMonitor':
      - keycloak
    '*v1.PrometheusRule':
      - keycloak
    '*v1.Secret':
      - credential-example-keycloak
      - keycloak-db-secret
  version: ''
```

Can we also set the `version` field?

pkg/common/cluster_state.go
```diff
@@ -349,11 +350,77 @@ func (i *ClusterState) readKeycloakIngressCurrentState(context context.Context,
...
+	postgresqlDeploymentReady, err := i.isDeploymentReady(i.PostgresqlDeployment)
+	if err != nil {
+		return false, err
+	}
+
+	// If running on OpenShift, check the Route is ready
+	stateManager := GetStateManager()
+	openshift, keyExists := stateManager.GetState(RouteKind).(bool)
+	if keyExists && openshift {
+		keycloakRouteReady = i.isRouteReady(i.KeycloakRoute)
```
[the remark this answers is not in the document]
davidffrench: unfortunately it doesn't look like it has anything. All an Ingress has in the status is a single field called 'loadBalancer' - https://godoc.org/k8s.io/api/networking/v1beta1#IngressStatus

his is just an array of IPs and hosts.

davidffrench:
@pb82 I would agree, I would like to be writing the version as well. unfortunately the only thing we list for the image is https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/constants.go#L13

pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -213,3 +217,51 @@ func watchSecondaryResource(c controller.Controller, resourceKind string, o runt
...
+	}
+
+	instance.Status.Ready = resourcesReady
+	instance.Status.Message = ""
+
+	// If resources are ready and we have not errored before now, we are in a reconciling phase
+	if resourcesReady {
+		instance.Status.Phase = v1alpha1.PhaseReconciling
+	} else {
+		instance.Status.Phase = v1alpha1.PhaseInitalising
```
golangcibot: PhaseInitalising not declared by package v1alpha1 (from `typecheck`)

davidffrench:
@maleck13 If you have time to look over this, that would be great. This should be used on every operator in my opinion.

pkg/common/cluster_actions.go
slaskawi: Using `reflect` is the only way I know...

pkg/common/cluster_state.go
```diff
@@ -349,11 +351,77 @@ func (i *ClusterState) readKeycloakIngressCurrentState(context context.Context,
...
+		if !apiErrors.IsNotFound(err) {
 			return err
 		}
 	} else {
 		i.KeycloakIngress = keycloakIngress.DeepCopy()
 	}
 	return nil
 }
+
+func (i *ClusterState) IsResourcesReady() (bool, error) {
```
slaskawi: I think it's the right point to think about structuring our model. At the moment, every part of our model needs to following functionality:
- Return a new structure
- Return a reconciliated structure (for `UpdateAction`)
- Return a selector to it (so that we can check if it has been created or not)
- and a new functionality you just introduced - checking if its ready. 

Each piece of the functionality is different for each model type (whether it's a Service, a Route or anything). Perhaps we should create a JIRA to restructure it?

pkg/common/cluster_actions.go
maleck13: not sure what the code is doing , you could also use ``` search.GetObjectKind().GroupVersionKind().Kind``` I believe

maleck13:
> @maleck13 If you have time to look over this, that would be great. This should be used on every operator in my opinion.

Is this the sub resource checking? If so I agree operators should be checking the sub resources and ensuring the values are as expected

pkg/apis/keycloak/v1alpha1/keycloak_types.go
pb82: What about specifying a tag in addition to the image [here](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/constants.go#L13)? Then we could use the tag as the version.

pkg/common/cluster_actions.go
davidffrench: Thanks @maleck13 . These methods do exist, unfortunately the Kind is an empty string. I believe because this is initial creation and is using our Golang objects with the type set but no specific Kind field.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
davidffrench: Yes, this is exactly what I have suggested on Sebastians PR https://github.com/keycloak/keycloak-operator/pull/43#discussion_r332582644

davidffrench:
@pb82 @slaskawi @abstractj Please let me know if there are any additional changes you think should be made to this PR.

PR #58, opened 2019-10-30
INTLY-3699 fix prometheus expressions and update grafana dashboard

pb82:
## JIRA ID

INTLY-3699

## Additional Information

This fixes the Prometheus rules so that they can be imported. Also fixes a problem where the Grafana dashboard caused constant reconciliation (because both operators, keycloak and grafana modified it). I've also updated to a better dashboard.

## Verification Steps

1. Deploy the application monitoring stack
2. Deploy keycloak (or rhsso) in another namespace (it must have the monitoring-key=middleware label)
3. Make sure the dashboard and prometheus rules are imported
4. There will be no data to display because that would require the metrics SPI


pkg/model/monitoring_constants.go
```diff
@@ -0,0 +1,1138 @@
+package model
+
+const GrafanaDashboardJson = `{
```
golangcibot: const `GrafanaDashboardJson` should be `GrafanaDashboardJSON` (from `golint`)

PR #65, opened 2019-11-05
INTLY-3235 External database support

slaskawi:
## JIRA ID

[INTLY-3235](https://issues.jboss.org/browse/INTLY-3235)

## Additional Information

This Pull Request introduces external Postgresql support for Keycloak Operator. It reuses the same `keycloak-db-secret` but requires two additional fields:
* `POSTGRES_EXTERNAL_ADDRESS`
* `POSTGRES_EXTERNAL_PORT`

## Verification Steps

This YAML might be useful for testing:

```yaml
apiVersion: v1
data:
  POSTGRES_DATABASE: cm9vdA==
  # This needs to be updated with the following value
  # oc get pods -o wide (and copy-paste Postgresql IP)
  POSTGRES_EXTERNAL_ADDRESS: MTcyLjE3LjAuMw==
  POSTGRES_EXTERNAL_PORT: NTQzMg==
  POSTGRES_HOST: a2V5Y2xvYWstcG9zdGdyZXNxbA==
  POSTGRES_PASSWORD: ZXhhbXBsZS1rZXljbG9hay1lTWFQ
  POSTGRES_SUPERUSER: dHJ1ZQ==
  POSTGRES_USERNAME: ZXhhbXBsZS1rZXljbG9hay1WbGdU
kind: Secret
metadata:
  labels:
    app: keycloak
  name: keycloak-db-secret
  namespace: keycloak
type: Opaque
---
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
  labels:
    app: keycloak
    component: database
  name: keycloak-postgresql
  namespace: keycloak
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: keycloak
      component: database
  strategy:
    type: Recreate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: keycloak
        component: database
      name: keycloak-postgresql
      namespace: keycloak
    spec:
      containers:
      - env:
        - name: POSTGRES_USER
          valueFrom:
            secretKeyRef:
              key: POSTGRES_USERNAME
              name: keycloak-db-secret
        - name: PGUSER
          valueFrom:
            secretKeyRef:
              key: POSTGRES_USERNAME
              name: keycloak-db-secret
        - name: POSTGRES_PASSWORD
          valueFrom:
            secretKeyRef:
              key: POSTGRES_PASSWORD
              name: keycloak-db-secret
        - name: POSTGRES_DB
          value: root
        - name: PGDATA
          value: /var/lib/postgresql/data/pgdata
        image: postgres:9.5
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 3
          initialDelaySeconds: 30
          periodSeconds: 10
          successThreshold: 1
          tcpSocket:
            port: 5432
          timeoutSeconds: 1
        name: keycloak-postgresql
        ports:
        - containerPort: 5432
          protocol: TCP
        readinessProbe:
          exec:
            command:
            - /bin/sh
            - -c
            - psql -h 127.0.0.1 -U $POSTGRES_USER -q -d $POSTGRES_DB -c 'SELECT 1'
          failureThreshold: 3
          initialDelaySeconds: 5
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
---
kind: Keycloak
metadata:
  labels:
    app: sso
  name: example-keycloak
  namespace: keycloak
spec:
  externalAccess:
    enabled: true
  externalDatabase:
    enabled: true
  instances: 1
```

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


docs/external-database.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+
+Keycloak Operator allows you to use an external Postgresql database by modifying `keycloak-postgresql` Endpoints object.
+The Operator also reuses the same `keycloak-db-secret` Secret but will require two additional properties:
+
+* POSTGRES_EXTERNAL_ADDRESS - an IP address (or a Hostname) or the external database. This address needs
+  to be resolvable from Kubernetes cluster perspective.
+* POSTGRES_EXTERNAL_PORT - (Optional) A database port.
+
+The other properties remain the same for both operating modes (hosted Postgresql and an external one).
+
```
davidffrench: @slaskawi Thank you for adding docs on this. It will be very useful to the end users of this operator.

I think this section should just first list what the user needs to provide to allow external database support and then if we still want to go into an explanation on how it is done internally in the operator.

```suggestion
By default, the operator will create a Postgres instance and database secret with each Keycloak CR. To enable external database support, you need to do two steps.

1. Create a Secret resource with the fields listed below
2. Specify the spec.externalDatabase.databaseSecretName in the Keycloak CR spec

 *  POSTGRES_DATABASE: //TODO
 *  POSTGRES_HOST:  //TODO
 *  POSTGRES_PASSWORD:  //TODO
*   POSTGRES_SUPERUSER:  //TODO
*   POSTGRES_USERNAME:  //TODO
* POSTGRES_EXTERNAL_ADDRESS - an IP address (or a Hostname) or the external database. This address needs
  to be resolvable from Kubernetes cluster perspective.
* POSTGRES_EXTERNAL_PORT - (Optional) A database port.

> Here's an example of such secret (note that values are Base64 encoded, see https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets[Kubernetes Secrets manual]:
> 
> .`keycloak-db-secret` Secret
> ```yaml
> apiVersion: v1
> data:
>   POSTGRES_DATABASE: cm9vdA==
>   POSTGRES_EXTERNAL_ADDRESS: MTcyLjE3LjAuMw==
>   POSTGRES_EXTERNAL_PORT: NTQzMg==
>   POSTGRES_HOST: a2V5Y2xvYWstcG9zdGdyZXNxbA==
>   POSTGRES_PASSWORD: ZXhhbXBsZS1rZXljbG9hay1lTWFQ
>   POSTGRES_SUPERUSER: dHJ1ZQ==
>   POSTGRES_USERNAME: ZXhhbXBsZS1rZXljbG9hay1WbGdU
> kind: Secret
> metadata:
>   labels:
>     app: keycloak
>   name: keycloak-db-secret
>   namespace: keycloak
> type: Opaque
> ```
> 
> The `Keycloak` Custom Resource also needs to be updated to turn the external database support on.
> Here's an example:
> 
> .`Keycloak` Custom Resource with external database support
> ```yaml
> apiVersion: keycloak.org/v1alpha1
> kind: Keycloak
> metadata:
>   labels:
>     app: sso
>   name: example-keycloak
>   namespace: keycloak
> spec:
>   externalDatabase:
>     enabled: true
>   instances: 1
> ```

```

Had to quote the copied examples due to backticks.

docs/external-database.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+Here's an example of such secret (note that values are Base64 encoded, see https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets[Kubernetes Secrets manual]:
+
+.`keycloak-db-secret` Secret
+```yaml
+apiVersion: v1
+data:
+  POSTGRES_DATABASE: cm9vdA==
+  POSTGRES_EXTERNAL_ADDRESS: MTcyLjE3LjAuMw==
+  POSTGRES_EXTERNAL_PORT: NTQzMg==
+  POSTGRES_HOST: a2V5Y2xvYWstcG9zdGdyZXNxbA==
```
davidffrench: It is unfortunate the end user needs to supply this as it is the service name that is still created by our operator.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -10,18 +10,22 @@ import (
...
+	ExternalDatabase      KeycloakExternalDatabase `json:"externalDatabase,omitempty"`
+	Profile               string                   `json:"profile,omitempty"`
 }
 
 type KeycloakExternalAccess struct {
 	Enabled bool `json:"enabled,omitempty"`
 }
 
+type KeycloakExternalDatabase struct {
+	Enabled bool `json:"enabled,omitempty"`
```
davidffrench: It would be good if we allowed the end user to supply a secret name here instead of the boolean. my concern is that the secret name needs to match what we currently call the secret in the operator. If this is left as is, it needs to be pointed out in the docs that the secret needs a specific name.

pkg/common/cluster_state.go
```diff
@@ -171,6 +177,22 @@ func (i *ClusterState) readPostgresqlServiceCurrentState(context context.Context
...
+	postgresqlServiceEndpointsSelector := model.PostgresqlServiceEndpointsSelector(cr)
+
+	err := controllerClient.Get(context, postgresqlServiceEndpointsSelector, postgresqlServiceEndpoints)
+	if err != nil {
+		if !apiErrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		i.PostgresqlServiceEndpoints = postgresqlServiceEndpoints.DeepCopy()
+		cr.UpdateStatusSecondaryResources(i.PostgresqlService.Kind, i.PostgresqlService.Name)
```
davidffrench:
```suggestion
		cr.UpdateStatusSecondaryResources(i.PostgresqlServiceEndpoints.Kind, i.PostgresqlServiceEndpoints.Name)
```

davidffrench: This is unfortunately also a problem as this Endpoints resource type will get added to the secondary resources in the status even when an external database is not added.  This is also a problem for the in the case where externalAccess is false for a route or an ingress. We should probably not be reading their state from the cluster in cases where we know we won't need them.

pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -283,3 +289,14 @@ func (i *KeycloakReconciler) getKeycloakIngressDesiredState(clusterState *common
 		Msg: "Update Keycloak Ingress",
 	}
 }
+
+func (i *KeycloakReconciler) getPostgresqlServiceEndpointsDesiredState(clusterState *common.ClusterState, cr *kc.Keycloak) common.ClusterAction {
+	if clusterState.PostgresqlServiceEndpoints == nil {
+		// This happens only during initial run
+		return nil
+	}
+	return common.GenericUpdateAction{
```
davidffrench: How come there is no create action here and only an update?

slaskawi: The `Endpoints` objects is created and maintained by Kubernetes in general. Typically, they shouldn't be touched by the users, however, redirecting traffic to an external thing is one of such use cases. 

You may find more information about this approach on OpenShift Documentation (1). However, the latest version of the documentation has been restructured and this page has been wiped out. 

(1) https://docs.openshift.com/container-platform/3.11/dev_guide/integrating_external_services.html

pkg/common/cluster_state.go
slaskawi: Good point. I'll remove it from the secondary resources.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
slaskawi: I've been talking about this on our internal channel. The conclusion was the we can either reference a secret or reuse the same we would create in hosted database mode. 

I decided to for the latter as the integration code is already there.

I also tried to be very explicit about the expected secret name in the docs. Could you please have a look at them and tell me, what parts could be improved?

docs/external-database.asciidoc
slaskawi: Actually `POSTGRES_HOST` is not needed by our Operator. This is needed by Integreately Backups :disappointed: See https://github.com/integr8ly/backup-container-image/blob/master/image/tools/lib/component/postgres.sh#L16

Personally, I would love to remove it but then, backups wouldn't work :disappointed:

docs/external-database.asciidoc
slaskawi: I restructured the docs. Please have a look once you have a chance.

pkg/common/cluster_state.go
slaskawi: Actually, I have a better idea! I'll add the Endpoints only if the external database support is turned on.

slaskawi:
@davidffrench Comments addressed and rebased.

docs/external-database.asciidoc
davidffrench: Thanks @slaskawi . What I was talking about here was that the HOST from what I can see in the code is the name of the database service we create - https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/database_secret.go#L53

This means the end user will need to know this secret name. I think for now it is fine to just explicitly list that in the docs.

docs/external-database.asciidoc
davidffrench: The docs are looking much clearer now in my opinion. Just on this line:
https://github.com/keycloak/keycloak-operator/pull/65/files#diff-11ba1e9a3a3933aaafa03e5e34f91569R32

Since this is the only time the user will be creating the database secret, I don't believe they will be familiar with the existing properties required. Would you agree or am I missing something here?

pkg/apis/keycloak/v1alpha1/keycloak_types.go
davidffrench: I agree, reusing the same secret is perfect and should definitely be done. Allowing them to specify a different name to this existing secret is more of an enhancement and not required for now. It is very clear in the docs, thank you

pkg/common/cluster_state.go
davidffrench: perfect now, thank you

pkg/controller/keycloak/keycloak_reconciler.go
davidffrench: ah ok, perfect. Thank you.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
pb82: I think there would still be benefit in having a configurable secret name. What about reading `DatabaseSecretName` from the CR (see e.g. https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/postgresql_deployment.go#L77) and set it to a default value of `<profile>-db-secret`?

davidffrench:
@slaskawi Unfortunately I am getting a runtime error on this line

https://github.com/keycloak/keycloak-operator/pull/65/files#diff-ccb60753eaccbee88e6bd4e51f01fc53R50

```
(runtime error: index out of range [0] with length 0)
```

Unfortunately, the subsets slice is empty on the reconciled object.

docs/external-database.asciidoc
slaskawi: ooh I get what you mean. Let me update the docs.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
slaskawi: I'm still not convinced to be honest. We allow only one Keycloak installation per namespace, so there can be only one database secret. I can't what could making the secret name configurable buy us. 

So I would propose to do the following - if you have in mind a usecase, where we need it - please let me know. I'll add it then. If you think more of a future and some possible usecases - let's just create a JIRA for it. This way we could implement it later. Would that work for you?

docs/external-database.asciidoc
slaskawi: Fair point. I added proper docs around it.

slaskawi:
> Unfortunately, the subsets slice is empty on the reconciled object.

@davidffrench That's a funny story. I had a check for it but it seemed, Kubernetes always created the first slice for me. On what version did you check this?

slaskawi:
Rebased and updated.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
davidffrench: +1 from me

docs/external-database.asciidoc
davidffrench: Thanks @slaskawi

davidffrench:
@slaskawi I am checking on K8s 1.11

docs/external-database.asciidoc
davidffrench: looks great

PR #67, opened 2019-11-06
INTLY-3223 KeycloakClient CR implementation

slaskawi:
## JIRA ID

[INTLY-3223](https://issues.jboss.org/browse/INTLY-3223)

## Additional Information

This Pull Request introduces support for Keycloak Client CRs.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

## Additional Notes

Still to be done:
* [x] Update CRD to match reality
* [x] Prepare a simple example with a public client
* [x] Prepare a simple example with a secret client
* [x] Implement some tests
* [x] Add documentation


pkg/model/constants.go
```diff
@@ -39,4 +39,7 @@ const (
 	KeycloakExtensionPath               = "/opt/jboss/keycloak/providers"
 	KeycloakExtensionsInitContainerPath = "/opt/extensions"
 	RhssoExtensionPath                  = "/opt/eap/providers"
+	ClientSecretName                    = ApplicationName + "-client-secret"
+	ClientSecretClientIdProperty        = "CLIENT_ID"
```
golangcibot: const `ClientSecretClientIdProperty` should be `ClientSecretClientIDProperty` (from `golint`)

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,294 @@
...
+	client   client.Client
+	scheme   *runtime.Scheme
+	context  context.Context
+	cancel   context.CancelFunc
+	recorder record.EventRecorder
+}
+
+// Reconcile reads that state of the cluster for a KeycloakClient object and makes changes based on the state read
+// and what is in the KeycloakClient.Spec
+func (r *ReconcileKeycloakClient) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
golangcibot: Function 'Reconcile' has too many statements (44 > 40) (from `funlen`)

deploy/crds/KeycloakClient.yaml
```diff
@@ -0,0 +1,221 @@
...
+              additionalProperties:
+                items:
+                  type: string
+                type: array
+              description: 'A map of all the secondary resources types and names created
+                for this CR. e.g "Deployment": [ "DeploymentName1", "DeploymentName2"
+                ]'
+              type: object
+          type: object
+      type: object
```
davidffrench:
```suggestion
```
Needs to be removed to work on k8s 1.11

deploy/examples/client/client-secret.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: keycloak.org/v1alpha1
+kind: KeycloakClient
+metadata:
+  name: client-secret
+spec:
+  realmSelector:
+    matchExact: master
+  instanceSelector:
```
davidffrench: seeing this beside the `realmSelector`, this might be better changed to `keycloakSekector`. However, since we are only targetting single namespace support for now, this can be removed and added later once multi namespace support is added. What do you think?

deploy/role.yaml
```diff
@@ -4,96 +4,97 @@ metadata:
...
+- apiGroups:
+  - keycloak.org
+  resources:
+  - keycloaks
+  - keycloaks/status
+  - keycloaks/finalizers
+  - keycloakrealms
+  - keycloakrealms/status
+  - keycloakrealms/finalizers
+  - keycloakclients
```
davidffrench:
```suggestion
  - keycloakclients
  - keycloakclients/status
  - keycloakclients/finalizers
```
The sub resources need to be listed here as well.

pkg/common/client_state.go
```diff
@@ -0,0 +1,60 @@
...
+	v1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ClientState struct {
+	Client       *kc.KeycloakAPIClient
+	ClientSecret *v1.Secret
+	Context      context.Context
+	Keycloak     *kc.Keycloak
```
davidffrench: I can't find this used anywhere. Should be able to remove unless I am missing something.

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	err = r.validate(instance)
+	if err != nil {
+		return r.ManageError(instance, err)
+	}
+
+	r.adjustCR(instance)
+
+	keycloaks, err := r.getMatchingKeycloaks(instance)
```
davidffrench: This could be switched to just the keycloak in the same namespace as the client creation or even to default to that if no `instanceSelector` is specified in the Spec

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	// If there's no client id, we can't process this any further
+	if instance.Spec.Client == nil {
+		return fmt.Errorf("client %v/%v has no Client definition and will be ignored", instance.Namespace, instance.Name)
+	}
+	if instance.Spec.Client.ClientID == "" {
+		return fmt.Errorf("client %v/%v has no Client ID and will be ignored", instance.Namespace, instance.Name)
+	}
+	return nil
+}
```
davidffrench: This can all be done using kubebuilder markers in the types file. The benefit of this is that the user should get an immediate response that they have specified something wrong in the Spec rather than success on the `create` and an async failure through the status.

see https://book.kubebuilder.io/reference/markers/crd-validation.html

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,306 @@
...
+		if err != nil {
+			return r.ManageError(instance, err)
+		}
+	}
+
+	return reconcile.Result{Requeue: false}, r.manageSuccess(instance, instance.DeletionTimestamp != nil)
+}
+
+// Fills the CR with default values. Nils are not acceptable for Kubernetes.
+func (r *ReconcileKeycloakClient) adjustCR(cr *kc.KeycloakClient) {
```
davidffrench:
```suggestion
func (r *ReconcileKeycloakClient) setCrDefaults(cr *kc.KeycloakClient) {
```
Would prefer a more explicit name

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	r.adjustCR(instance)
+
+	keycloaks, err := r.getMatchingKeycloaks(instance)
+	if err != nil {
+		return r.ManageError(instance, err)
+	}
+
+	log.Info(fmt.Sprintf("found %v matching keycloak(s) for realm %v/%v", len(keycloaks.Items), instance.Namespace, instance.Name))
+
+	// The realm may be applicable to multiple keycloak instances,
```
davidffrench:
```suggestion
	// The client may be applicable to multiple keycloak instances,
```

deploy/crds/KeycloakClient.yaml
slaskawi: Fixed.

deploy/examples/client/client-secret.yaml
slaskawi: I would actually leave it as is. I think the multi-namespace support will become a priority very soon.

pkg/controller/keycloakclient/keycloakclient_controller.go
slaskawi: Fixed.

deploy/role.yaml
slaskawi: Fixed.

pkg/controller/keycloakclient/keycloakclient_controller.go
slaskawi: Added `//nolint`

pkg/common/client_state.go
slaskawi: Fixed.

pkg/controller/keycloakclient/keycloakclient_controller.go
slaskawi: I agree to that. I think we could always assume that whenever `instanceSelector` is `nil`, we target the same namespace. However, the same mechanics needs to be used for `KeycloakRealm`, `KeycloakClient` and `KeycloakUser`. And since 2 out of 3 are still in a PR state, I believe we should implement it as soon as all the code is merged. Does it makes sense? Otherwise I worry we'll end up with 3 different solutions to the same problem.

pkg/controller/keycloakclient/keycloakclient_controller.go
slaskawi: Good catch! Fixed.

deploy/examples/client/client-secret.yaml
davidffrench: sounds good to me

pkg/controller/keycloakclient/keycloakclient_controller.go
davidffrench: ya, makes perfect sense. No problem with this being in a separate PR

pkg/controller/keycloakclient/keycloakclient_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	// be added to. Skip reconcile until a selector has been set.
+	if instance.Spec.InstanceSelector == nil {
+		return fmt.Errorf("realm %v/%v has no instance selector and will be ignored", instance.Namespace, instance.Name)
+	}
+	// If no realm selector is set we can't figure out which Keycloak Client we need to target for the API calls
+	if instance.Spec.RealmSelector == nil {
+		return fmt.Errorf("client %v/%v has no realm selector and will be ignored", instance.Namespace, instance.Name)
+	}
+	return nil
+}
```
davidffrench: @slaskawi Can these be set as kubebuilder validation in the CR spec as well?

slaskawi: Done

docs/keycloak-client.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+  client:
+    # auto-generated if not supplied
+    #id: 123
+    clientId: client-secret
+    secret: client-secret
+    # ...
+    # other properties of Keycloak Client
+```
+
+TIP: Note, that `instanceSelector` needs to match labels of an existing `Keycloak` Custom Resource.
```
davidffrench:
```suggestion
```

slaskawi: Fixed. Thanks!

PR #73, opened 2019-11-08
feat: Upgrade Postgres to 11.5

davidffrench:
## JIRA ID
[KEYCLOAK-12259](https://issues.jboss.org/browse/KEYCLOAK-12259)

## Additional Information
Based on a mail thread, Postgres should be updated to 10.1 but this is not working with OpenShift due to permissions. It is working with 11.5 but we need confirmation we can upgrade to this.

## Verification Steps
1. Create Keycloak instance from this branch
2. Run any other tests you think are required.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/model/constants.go
+++ b/pkg/model/constants.go
@@ -10,7 +10,7 @@ const (
 	PostgresqlDeploymentName             = ApplicationName + "-postgresql"
 	PostgresqlDeploymentComponent        = "database"
 	PostgresqlServiceName                = ApplicationName + "-postgresql"
-	PostgresqlImage                      = "postgres:9.5"
+	PostgresqlImage                      = "postgres:11.5"
 	KeycloakImage                        = "quay.io/keycloak/keycloak:7.0.1"
 	KeycloakInitContainerImage           = "quay.io/keycloak/extensions-init-container:master"
 	RHSSOImage                           = "registry.access.redhat.com/redhat-sso-7/sso73-openshift:1.0-15"


slaskawi:
Please do not merge it just now. I plan to test it out (with a few migration scenarios) today or tomorrow.

stianst:
What level of testing do we have around DBs? We do really need to automate all testing and not require manual testing to update the DB version.

slaskawi:
I've tested this one locally and I guess it's fine to merge it. There's one thing to consider though - if you're upgrading existing operator installation - you might lose some data. 

It seems when upgrading the Postgresql it wipes out everything from the DB. Once we upgrade Keycloak image (which will happen very soon), the migration addon will restart Keycloak and everything will eventually be populated to the DB. However, you will lose all manual modification you did. The only solution for that is to do a backup of your data prior to the upgrade.

slaskawi:
> What level of testing do we have around DBs? We do really need to automate all testing and not require manual testing to update the DB version.

@stianst We got extremely unlucky with this one. Just before the initial release, we were told that the OpenShift Team is not planning to maintain Postgresql `9.5` in `4.x` series. Even though that came as a surprise - we're using the community image, which is in perfect health. So this time we're lucky I guess but once the image gets productized, we need to work on automatic upgrades that include database version or type migration (basically [KEYCLOAK-12027](https://issues.jboss.org/browse/KEYCLOAK-12027)).

PR #77, opened 2019-11-12
Adjust timeouts for readiness and liveness probes

slaskawi:
## Additional Information

This PR tunes readiness and liveness probes to me more responsive and tolerant. It might be merged after the initial release.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/model/keycloak_deployment.go
+++ b/pkg/model/keycloak_deployment.go
@@ -146,8 +146,9 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
 							},
 							VolumeMounts: KeycloakVolumeMounts(KeycloakExtensionPath),
 							LivenessProbe: &v1.Probe{
-								InitialDelaySeconds: 60,
+								InitialDelaySeconds: 30,
 								TimeoutSeconds:      1,
+								FailureThreshold:    20,
 								Handler: v1.Handler{
 									HTTPGet: &v1.HTTPGetAction{
 										Path:   "/auth/realms/master",
@@ -157,8 +158,9 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
 								},
 							},
 							ReadinessProbe: &v1.Probe{
+								InitialDelaySeconds: 30,
 								TimeoutSeconds:      1,
-								InitialDelaySeconds: 10,
+								FailureThreshold:    10,
 								Handler: v1.Handler{
 									HTTPGet: &v1.HTTPGetAction{
 										Path:   "/auth/realms/master",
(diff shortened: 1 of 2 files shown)


pkg/model/keycloak_deployment.go
```diff
@@ -146,8 +146,9 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
 							},
 							VolumeMounts: KeycloakVolumeMounts(KeycloakExtensionPath),
 							LivenessProbe: &v1.Probe{
-								InitialDelaySeconds: 60,
+								InitialDelaySeconds: 30,
 								TimeoutSeconds:      1,
+								FailureThreshold:    20,
```
stianst: FailureThreshold 20 seems very high for a liveness probe.

pkg/model/keycloak_deployment.go
```diff
@@ -157,8 +158,9 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
 								},
 							},
 							ReadinessProbe: &v1.Probe{
+								InitialDelaySeconds: 30,
```
stianst: 30 for initial delay seems a bit high. On fast machines Keycloak can start in <15 seconds.

pkg/model/keycloak_deployment.go
```diff
@@ -157,8 +158,9 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak) *v13.StatefulSet {
...
 								TimeoutSeconds:      1,
-								InitialDelaySeconds: 10,
+								FailureThreshold:    10,
```
stianst: Failure threshold 10 should be ok in most cases, but what happens when we're doing an upgrade and migration code slows startup down?

stianst:
Now that we have a release out we need to have a JIRA for all changes, so please create one.

pkg/model/keycloak_deployment.go
slaskawi: I'm still thinking how to implement this...

There are two very problematic usecases (out of three). 
1. The easiest one is when you have a descent hardware and everything is up and running within 15 seconds. 
2. The second, problematic, is Travis. Sometimes it happens that the builds take much, much longer and usual. It's not very often, but I've seen it 2-3 times for the whole our development cycle. I've seen timeouts in 90 seconds (60 for the initial delay plus 3 times 10 second for standard probe settings). 
3. The worst case are the demos - CRC (or even Minikube) running with BlueJeans. In this case you can see timeouts even with 2-3 mins. 

So I'm scratching my head, how to cover all those 3 use cases. The best idea that I have is to use fairly small `InitialDelaySeconds` so that Kubernetes starts to probe it as soon as possible (this covers no. 1). Then we should probe pretty often checking if it's ready/alive with 10s in-between (as it can come up anytime). To make everything work we need allow multiple probe failures.

I'm still not convinced it will work. I need to do more testing around it (with different CPU settings on my Minikube). We also have https://github.com/keycloak/keycloak-operator/issues/95. So maybe we should set the default as good as we can and then don't do any reconciliation on it (letting users override it).

stianst: Just to add one more. If you expect it to start quickly, but there's some error (like DB connection issue) you want it to fail fast.

stianst: That's ReadinessProbe though right. The LivenessProve starts after the ReadinessProbe is finished doesn't it?

PR #80, opened 2019-11-13
Make requeue constants actually seconds

grdryn:
Before this change, there were cases where nanoseconds were used.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/controller/keycloak/keycloak_controller.go
+++ b/pkg/controller/keycloak/keycloak_controller.go
@@ -36,8 +36,8 @@ import (
 var log = logf.Log.WithName("controller_keycloak")
 
 const (
-	RequeueDelaySeconds      = 30
-	RequeueDelayErrorSeconds = 5
+	RequeueDelaySeconds      = 30 * time.Second
+	RequeueDelayErrorSeconds = 5 * time.Second
 	ControllerName           = "keycloak-controller"
 )
 
@@ -247,5 +247,5 @@ func (r *ReconcileKeycloak) ManageSuccess(instance *v1alpha1.Keycloak, currentSt
 	}
 
 	log.Info("desired cluster state met")
-	return reconcile.Result{RequeueAfter: RequeueDelaySeconds * time.Second}, nil
+	return reconcile.Result{RequeueAfter: RequeueDelaySeconds}, nil
 }


pb82:
merged, thanks @grdryn

pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -36,8 +36,8 @@ import (
 var log = logf.Log.WithName("controller_keycloak")
 
 const (
-	RequeueDelaySeconds      = 30
-	RequeueDelayErrorSeconds = 5
+	RequeueDelaySeconds      = 30 * time.Second
```
golangcibot: var RequeueDelaySeconds is of type time.Duration; don't use unit-specific suffix "Seconds" (from `golint`)

pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -36,8 +36,8 @@ import (
...
+	RequeueDelayErrorSeconds = 5 * time.Second
```
golangcibot: var RequeueDelayErrorSeconds is of type time.Duration; don't use unit-specific suffix "Seconds" (from `golint`)

pkg/controller/keycloak/keycloak_controller.go
grdryn: huh

slaskawi: hmmmm that's weird. So it worked correctly before? Even without this change?

grdryn: I guess it's because it's now a `time.Duration` where before it was an `int64`?

slaskawi: You tell me :) Jokes aside - please check it locally and if needed, create a Pull Request with a revert.

grdryn: Oh yeah, this part works fine, it's just a lint warning. Where this value was previously used it was being converted to a `time.Duration`, but I guess golint has a rule to warn against appending a "Seconds" suffix to values of `time.Duration`.

Now, speaking of running locally, that didn't work for a different reason: either `https://quay.io/keycloak/extensions-init-container` doesn't exist or it's a private repo.

slaskawi: Yes, we have a PR to fix that: https://github.com/keycloak/keycloak-operator/pull/81

PR #89, opened 2019-11-14
[PROPOSAL] Enhancements to the Makefile and Docker

abstractj:
## PROPOSAL
This PR is **not** intended to be merged before the first release of the new Operator and the goal is to provide small enhancements to overcome issues with cache on quay.io or during development. 

It was split into 3 separate commits, so if there's any disagreement we can just get rid of some of them or all of them :)

## Additional Information
* Removal of the previous approach to overcome issues with caching
* Enhancements to the Makefile

    When building the Docker image with Podman or on quay.io the shell script
    will be cached as we found out during the development.

    This kind of situation is not obvious for this reason was added:
    - docker/build: builds the Docker image and tag it for developers willing to use it locally
    - docker/release/prepare: to workaround the caching issues on quay.io Sebastian found out that we need to update an
    env variable at each release, until quay.io, fixes it or provides alternatives to disable cache.

* Small enhancements to the build

    The goal of this commit is to provide developers some guidance when struggling with caching. By adding these flags, people are able to know exactly which commit id the Keycloak Operator was built and when.
   Also, the Version of the Operator was moved to the Makefile, so we can
reuse it to build our Docker image

## Verification Steps
* Run `make docker/build` several times and verify that nothing is being cached
* Run `make docker/release/prepare` and verify the commit message: `Prepare for release version: 7.0.1 (git+sha: cd55b24, built: 20191114_215444Z)`

## Notes
* Kudos to @slaskawi  for finding the culprit of the issue
* This is not a definitive fix, it's just a proposal to not go crazy during the development
* Ideally quay.io should provide a way to disable cache

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,14 +9,10 @@ ENV OPERATOR=/usr/local/bin/keycloak-operator \
     USER_UID=1001 \
     USER_NAME=keycloak-operator
 
-ENV SHA1_FILE=/BUILD_SHA1.txt
-ENV BUILD_DATE_FILE=/BUILD_DATE.txt
 # Because of https://github.com/containers/buildah/issues/1906
 # We need to increment this variable whenever we need a rebuild.
 # The fix should be implemented very soon
 ENV CACHE_BUST=1
-
-RUN echo "$(date)" > $BUILD_DATE_FILE
 
 COPY tools /opt/jboss/tools
 RUN /opt/jboss/tools/build-operator.sh
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,6 @@
 package version
 
+//TODO Check if we can remove it
 var (
 	Version = "7.0.1"
 )
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,15 @@
 # Other contants
 NAMESPACE=keycloak
+VERSION=7.0.1
 PROJECT=keycloak-operator
 PKG=github.com/keycloak/keycloak-operator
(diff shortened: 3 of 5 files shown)


build/Dockerfile
```diff
@@ -9,15 +9,11 @@ ENV OPERATOR=/usr/local/bin/keycloak-operator \
     USER_UID=1001 \
     USER_NAME=keycloak-operator
 
-ENV SHA1_FILE=/BUILD_SHA1.txt
```
slaskawi: To be honest, I liked the approach with those 2 files. It's very convenient, you can `ssh` (or `rsh`) to a running container and find out what version your on in just a second.

version/version.go
```diff
@@ -1,5 +1,6 @@
 package version
 
+//TODO Check if we can remove it
 var (
 	Version = "7.0.1"
```
slaskawi: I think @davidffrench wanted to use it somewhere?

davidffrench: ya, i wanted it in the status, the version field is currently blank but I don't mind where it comes from. Also, there will be an enhancement soon to expose the version through the operator metrics endpoint as a new metric. Again, as long as it can be pulled from somewhere.

abstractj:
@davidffrench @sebastienblanc be honest we don't need to change anything, if you guys see something as useful I update the PR, otherwise, there's absolutely np about closing it.

abstractj:
Closing it, because we didn't reach an agreement on whether or not this is helpful.

PR #94, opened 2019-11-21
fix realm overrides

pb82:
Bugfix: the realm override for a browser redirect used the wrong property for setting the default provider.


pkg/common/cluster_actions.go
```diff
@@ -257,7 +257,7 @@ func (i *ClusterActionRunner) configureBrowserRedirector(provider, flow string,
 	if authenticatorConfig == nil && flow != "" {
```
davidffrench: Should this conditional also be checking for `provider` instead of `flow`?

pb82: yes, makes sense

PR #96, opened 2019-11-25
KEYCLOAK-12054 Use multi-stage builds

slaskawi:
## JIRA ID

[KEYCLOAK-12054](https://issues.jboss.org/browse/KEYCLOAK-12054)

## Additional Information

This Pull Request fixes cache issue on Quay (it was reusing image build cache for shell files, see https://github.com/containers/buildah/issues/1906). The proposed solution uses Docker Multi-Stage builds feature, where we have a builder image (that builds our Go program) and a final image (that contains a binary).

This approach has a lot of advantages:
- It makes the image smaller - from ~250 MB to ~54 MB (see [link1](https://quay.io/repository/slaskawi/keycloak-operator-private?tab=tags) [link2](https://quay.io/repository/keycloak/keycloak-operator?tab=tags)).
- It's just a pure UBI8 image with zero other dependencies, which means smaller CVE surface.
- This approach fixes `operator-hub build` command.
- It allows to play with build cache using `--cache-from` (so that you can reuse previous build results and modify only the final image).

However, there are some downsides as well:
- As it uses some newer Docker features, so "bye bye Docker `1.13`", which is installed by default in Fedora. From now on, please use Podman (or newer Docker version).
- I had to move the `Dockerfile` to the root and create a symbolic link to it from `build` directory.
- We need to change our build triggers to point to a root `Dockerfile` (Quay do not discover symbolic links correctly).

But overall, I think it's worth it. 

## Verification Steps

1. `podman build .` executed from project root
2. `operator-sdk-local build --image-builder podman whatever --verbose`
3. Run it in a cluster

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


pb82:
:eyes:

PR #113, opened 2019-12-16
KEYCLOAK-12409 - Makes image configurable

cmattoon:
Our use case is a fork of the specified image that includes custom themes, etc.

## JIRA ID
  * https://issues.redhat.com/browse/KEYCLOAK-12409
  * https://issues.redhat.com/browse/KEYCLOAK-12410

## Additional Information
I found issue #12409 after re-working some of this code, and so instead of `imageOverrides`, I opted to mirror the standard Kubernetes `container` struct - currently with `container.image` only - in anticipation of extra environment vars (see 12397), commands, etc.

## Verification Steps

Tested against a CRD like the following:
```
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  extensions:
    - https://github.com/aerogear/keycloak-metrics-spi/releases/download/1.0.4/keycloak-metrics-spi-1.0.4.jar
  externalAccess:
    enabled: True
  container:
    image: 1234567890.amazonaws.com/my-custom-image:tag
  initContainer:
    image: 1234567890.amazonaws.com/my-custom-init-container-image:tag
```

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloakbackups_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakbackups_crd.yaml
@@ -69,6 +69,13 @@ spec:
                 flag to true. Potentially, it will be possible to restore a single
                 backup multiple times."
               type: boolean
+            container:
+              type: object
+              description: Additional container configuration
+              properties:
+                image:
+                  description: The full Docker tag of the backup image to use
+                  type: string
           type: object
         status:
           description: KeycloakBackupStatus defines the observed state of KeycloakBackup
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -51,7 +51,7 @@ type KeycloakSpec struct {
 	Profile string `json:"profile,omitempty"`
 	// Image is the full docker tag of the keycloak image to use
 	Container KeycloakContainer `json:"container,omitempty"`
-	// InitContainer has informtion about the image tag to use
+	// InitContainer has information about the image tag to use
 	InitContainer KeycloakContainer `json:"initContainer,omitempty"`
 }
 
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -49,6 +49,14 @@ type KeycloakSpec struct {
 	// Profile used for controlling Operator behavior. Default is empty.
 	// +optional
 	Profile string `json:"profile,omitempty"`
+	// Image is the full docker tag of the keycloak image to use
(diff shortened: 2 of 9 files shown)


pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -49,6 +49,14 @@ type KeycloakSpec struct {
 	// Profile used for controlling Operator behavior. Default is empty.
 	// +optional
 	Profile string `json:"profile,omitempty"`
+	// Image is the full docker tag of the keycloak image to use
+	Container KeycloakContainer `json:"container,omitempty"`
+	// InitContainer has informtion about the image tag to use
```
golangcibot: `informtion` is a misspelling of `information` (from `misspell`)
```suggestion

```

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
```diff
@@ -69,6 +69,13 @@ spec:
                 flag to true. Potentially, it will be possible to restore a single
                 backup multiple times."
               type: boolean
+            container:
```
slaskawi: The Operator uses 4 images at the moment:
- a database image - Postgresql
- an Integreately Backup container
- a Keycloak image
- an Init image

The proposed implementation seems to support only Keycloak, which is not enough. We should allow to override all the images.

The `Container` struct doesn't look like a good fit as it is defined as follows:

```go
type Container struct {
   Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
   Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
}
```

So there's no obvious way to override any of the 4 images. 

In the [Keycloak Dev discussion](https://groups.google.com/forum/#!topic/keycloak-dev/OvgA_CNfgok) (linked from [KEYCLOAK-12409](https://issues.redhat.com/browse/KEYCLOAK-12409)) I proposed something like this:

```yaml
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  # Naming needs to be consistent with realmOverrides from KeycloakRealmCR
  # All fields are optional
  imageOverrides:
    operator: <custom image>
    postgresql <custom image>
    backup: <custom image>
``` 

I think it may work better as all images have a corresponding field (however, the example above misses init container and Keycloak images). Also, `imageOverrides` is similar to `realmOverrides` we already used in the code.

What do you think about this idea? Would you like to implement this changes?

abstractj:
@cmattoon could you please check @slaskawi's comments?

slaskawi:
@abstractj @cmattoon I believe this one can be closed. Depending on whoever will have time (@cmattoon or me), we'll provide a different PR with the implementation.

PR #120, opened 2020-01-15
KEYCLOAK-12669 Add PodDisruptionBudget support

wei-lee:
## JIRA ID

[KEYCLOAK-12669](https://issues.redhat.com/browse/KEYCLOAK-12669)

## Additional Information

Discussions about the topic can be found [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/keycloak-dev/A8F-sERqCGE/TmpO6p2MBwAJ)

## Verification Steps
1. Deploy the operator. You can use this image: `quay.io/wei_lee/keycloak-operator:dev`
2. Create a Keycloak CR. You can use the example one in this PR. At this point, there should be no PDB created yet as there is only 1 instance of Keycloak is provisioned.
3. Increase the number of Keycloak instances and you should see a PDB is created for Keycloak
4. Decrease the number of Keycloak to 1 again and you should see the PDB is removed.

## Checklist:
- [ ] Verified by team member
- [x] Automated Tests


wei-lee:
@slaskawi As we have discussed, this PR will add support for PDB. One thing that is slightly different from what we have discussed is that the operator will also take the number of instances of KC into consideration when creating the PDB. 

Please take a look and see what you think. I also have a cluster where I have this PR running and I can give the access to play with it if you want, just ping me on google chat and I can give you the credentials.

wei-lee:
@slaskawi I think if there is only 1 instance of Keycloak and the PDB is set to `maxUnavailable=1`, then  when you try to drain a node, it will always fail because it will violate the rule (there are no other replicas running anywhere else). If there is a timeout value set when draining the node, then it will eventually continue on. But if there is no timeout value set, it could block the node draining process.

@davidffrench can you confirm if I am right here?

wei-lee:
@slaskawi for some reason I can't merge the PR. It still says requires 1 approving review. If you can help merge the PR that will be great!

Thanks.

davidffrench:
> @slaskawi I think if there is only 1 instance of Keycloak and the PDB is set to `maxUnavailable=1`, then when you try to drain a node, it will always fail because it will violate the rule (there are no other replicas running anywhere else). If there is a timeout value set when draining the node, then it will eventually continue on. But if there is no timeout value set, it could block the node draining process.
> 
> @davidffrench can you confirm if I am right here?

Now that I am thinking about it, when using `maxUnavailable` set to 1 with 1 replica it will not block the node drain. When attempting to evict the pod, it will check if there is >= 1 pod in a non-ready state. In this case, that is false so it will evict the pod. 

If using `minAvailable` of 1 with 1 replica it would violate the PDB and would not evict the pod.

slaskawi:
> @slaskawi for some reason I can't merge the PR. It still says requires 1 approving review. If you can help merge the PR that will be great!

@wei-lee @stianst and @abstractj have merging rights in this repo.

I'm thinking about some improvements to this PR:
- at first we could use `maxUnavailable = 1` as @davidffrench suggested. This should always work fine right?
- If the above won't work - we could always create a PDB and always perform an update to it. If we have number of replicas equal to 1, we could set `minAvailable = 0`. This way we will be always creating a PDB and we'll get rid of this deleting PDB logic (which I don't like - it's not used anywhere else in our Operator). 

Let me take my approval back (sorry @wei-lee). Could you please address those comments?

wei-lee:
@slaskawi sounds good. I will make the change.

wei-lee:
@slaskawi I have made the changes as you suggested. Do you mind reviewing it again? Thanks.

pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -325,3 +325,36 @@ func (i *KeycloakReconciler) getPostgresqlServiceEndpointsDesiredState(clusterSt
...
+
+func (i *KeycloakReconciler) getPodDisruptionBudgetDesiredState(clusterState *common.ClusterState, cr *kc.Keycloak) common.ClusterAction {
+	if shouldHavePDB(cr) && clusterState.PodDisruptionBudget == nil {
+		return common.GenericCreateAction{
+			Ref: model.PodDisruptionBudget(cr),
+			Msg: "Create PodDisruptionBudget",
+		}
+	}
+
+	if shouldHavePDB(cr) && clusterState.PodDisruptionBudget != nil {
```
slaskawi: `shouldHavePDB(cr)` seems unnecessary here. We perform the same action if it's `true` and `false`. I believe it can be removed altogether (the `shouldHavePDB(cr)` method).

wei-lee: @slaskawi Hmm, I think `shouldHavePDB` is still needed otherwise I wouldn't be able to know what value the PDB object should be updated to. Depending on the value returned by `shouldHavePDB`, the `Ref` value is different on line L339 and L346.

slaskawi: aaah right... I didn't notice that. I'm not sure if we need any logic using `MaxAvailable` or `MinAvailable` at all...

What if we used always `PodDisruptionBudgetReconciled` - the one with `MaxUnavailable`? If my understanding is correct and I didn't miss anything, this one should always work (even if `cr.Spec.Instances == 1`). If my statement is true, I think we could get rid of both `shouldHavePDB` (and rely solely on `cr.Spec.PodDisruptionBudget.Enabled`). 

One thing might also deserve some testing - what will happen, if `cr.Spec.PodDisruptionBudget.Enabled`, we used only PDB's `MaxUnavailable` and `cr.Spec.Instances == 0`. This is a valid case we use during the migration (there's a short period of time, where we ask StatefulSet to kill all the nodes). 

What do you think?

wei-lee: yes, you are right, now I think about it, we can always use `PodDisruptionBudgetReconciled`, it will work regardless what number of the keycloak instances is set to. I will make the change as you suggested.

> One thing might also deserve some testing - what will happen, if cr.Spec.PodDisruptionBudget.Enabled, we used only PDB's MaxUnavailable and cr.Spec.Instances == 0. This is a valid case we use during the migration (there's a short period of time, where we ask StatefulSet to kill all the nodes).

In this [doc](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/), it specifically called out about PDB and rolling upgrades:

**_"Pods which are deleted or unavailable due to a rolling upgrade to an application do count against the disruption budget, but controllers (like deployment and stateful-set) are not limited by PDBs when doing rolling upgrades – the handling of failures during application updates is configured in the controller spec. (Learn about updating a deployment.)"_**

We have tested this as well and the presents of the PDB won't stop the pods from being scaled down during upgrades.

wei-lee:
@slaskawi I have updated the PR based on our discussion, mind reviewing it again? Thanks a lot!

pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -325,3 +325,21 @@ func (i *KeycloakReconciler) getPostgresqlServiceEndpointsDesiredState(clusterSt
...
 }
+
+func (i *KeycloakReconciler) getPodDisruptionBudgetDesiredState(clusterState *common.ClusterState, cr *kc.Keycloak) common.ClusterAction {
+	if cr.Spec.PodDisruptionBudget.Enabled {
+		if clusterState.PodDisruptionBudget == nil {
+			return common.GenericCreateAction{
+				Ref: model.PodDisruptionBudget(cr),
+				Msg: "Create PodDisruptionBudget",
+			}
+		} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

wei-lee:
@slaskawi mind reviewing this again? Thanks.

wei-lee:
Thanks for the help @slaskawi.

pkg/apis/keycloak/v1alpha1/zz_generated.openapi.go
```diff
@@ -11,22 +11,22 @@ import (
...
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakRealmStatus":  schema_pkg_apis_keycloak_v1alpha1_KeycloakRealmStatus(ref),
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakSpec":         schema_pkg_apis_keycloak_v1alpha1_KeycloakSpec(ref),
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakStatus":       schema_pkg_apis_keycloak_v1alpha1_KeycloakStatus(ref),
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakUser":         schema_pkg_apis_keycloak_v1alpha1_KeycloakUser(ref),
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakUserSpec":     schema_pkg_apis_keycloak_v1alpha1_KeycloakUserSpec(ref),
-		"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1.KeycloakUserStatus":   schema_pkg_apis_keycloak_v1alpha1_KeycloakUserStatus(ref),
+		"github.com/keycloak/keycloak-operator-1/pkg/apis/keycloak/v1alpha1.Keycloak":             schema_pkg_apis_keycloak_v1alpha1_Keycloak(ref),
+		"github.com/keycloak/keycloak-operator-1/pkg/apis/keycloak/v1alpha1.KeycloakAWSSpec":      schema_pkg_apis_keycloak_v1alpha1_KeycloakAWSSpec(ref),
```
JiyeYu: Hi @wei-lee,
Nice patch! Here is a bit of mix-up.  I guess the keycloak-operator-1 here is your local folder's name? 😄

stianst:
Please squash to a single commit, and prefix the commit message with the JIRA ("KEYCLOAK-12669 Add PodDisruptionBudget support").

stianst:
Please squash to a single commit, and prefix the commit message with the JIRA ("KEYCLOAK-12669 Add PodDisruptionBudget support").

stianst:
@davidffrench if this looks good to you know please change your review to approve

@wei-lee can you address last comment from @JiyeYu? and the comments from me?

davidffrench:
Thanks for your review @stianst and @JiyeYu . Wei is in and out of off-site meetings today so he may not respond immediately.

wei-lee:
thanks for the review.  I will get those issues addressed ASAP.

wei-lee:
@stianst @davidffrench comments addressed. Ready for merge.

PR #121, opened 2020-01-17
KEYCLOAK-12807 add options for audit logging

pb82:
## JIRA ID
https://issues.redhat.com/browse/INTLY-4404 (converted to Keycloak JIRA)
https://issues.redhat.com/browse/KEYCLOAK-12807

## Additional Information
Enable audit logging by default for realms created by the operator. This will enable recording of Admin Events, Login Events and JSON representations for Admin Events.

The list of events will look like this:

screenshot

## Verification Steps

1. Deploy keycloak using the operator from this branch
1. Create a realm and don't specify any additional parameters in the CR
1. Open the realm in the admin console and navigate to `Events > Config`
1. Both Login and Admin Events should be enabled and Include Representation should be enabled for Admin Events.

## Checklist:
- [ ] Verified by team member
- [x] Automated Tests

### DIFF
--- a/pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
@@ -43,6 +43,18 @@ type KeycloakAPIRealm struct {
 	// A set of Event Listeners.
 	// +optional
 	EventsListeners []string `json:"eventsListeners,omitempty"`
+	// Enable events recording
+	// TODO: change to values and use kubebuilder default annotation once supported
+	// +optional
+	EventsEnabled *bool `json:"eventsEnabled,omitempty"`
+	// Enable events recording
+	// TODO: change to values and use kubebuilder default annotation once supported
+	// +optional
+	AdminEventsEnabled *bool `json:"adminEventsEnabled,omitempty"`
+	// Enable admin events details
+	// TODO: change to values and use kubebuilder default annotation once supported
+	// +optional
+	AdminEventsDetailsEnabled *bool `json:"adminEventsDetailsEnabled,omitempty"`
 }
 
 type RedirectorIdentityProviderOverride struct {
--- a/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
@@ -60,8 +60,8 @@ spec:
                           type: string
                         type: array
                     required:
-                    - key
-                    - operator
+                      - key
+                      - operator
                     type: object
                   type: array
                 matchLabels:
@@ -215,7 +215,7 @@ spec:
                           type: string
                         type: array
                     required:
-                    - clientId
+                      - clientId
                     type: object
                   type: array
                 displayName:
@@ -359,7 +359,7 @@ spec:
                     type: object
                   type: array
               required:
-              - realm
+                - realm
               type: object
             realmOverrides:
               description: A list of overrides to the default Realm behavior.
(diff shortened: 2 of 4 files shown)


pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -43,6 +43,15 @@ type KeycloakAPIRealm struct {
 	// A set of Event Listeners.
 	// +optional
 	EventsListeners []string `json:"eventsListeners,omitempty"`
+	// Enable events recording
+	// +optional
+	EventsEnabled *bool `json:"eventsEnabled,omitempty"`
```
slaskawi: In this particular case, we use pointers since we want to enable this by default (and with values, we would get `false` here if a user hasn't specified anything). 

Perhaps we could use `// +kubebuilder:validation:Default=true` ? I think the shortcut should also work: `+kubebuilder:default=true`

See [this Pull Request](https://github.com/kubernetes-sigs/controller-tools/pull/323/files#diff-862d6e06a6705fcf1b7782d96f1e003aR98) for more info.

pb82: Interesting, didn't know about that. Will update the PR shortly.

pb82: @slaskawi @davidffrench As discussed, kubebuilder annotations do not seem to be supported in any operator-sdk release yet (but there seem to be plans). Leaving the code as is for now (because not sure if we should add a default value annotation for a pointer).

davidffrench: Thanks @pb82 . This was working for me before, but since the default field is only supported in CustomResourceDefinition v1, it was possibly removed from the CRD generation for now.

slaskawi: That's very interesting @davidffrench! Can we somehow ask `Operator SDK generate`  to produce CRD v1 for us? From our perspective - that would be a desirable move!

davidffrench: v1 is only available from k8s 1.16/OpenShift 4.3 . MY assumption is it is on their radar and will be available in a future release. It is already supported in controller-tools which does the actual generation.

slaskawi: @pb82 @davidffrench Ok, that sounds good to me. @pb82 May I ask for one additional, tiny change - would you mind to add a `//TODO` there saying that we may want to change it to values (instead of pointers) some day? Apart from that - it looks good to me.

pb82: @slaskawi Sure, done

jessesarn:
CS-SRE asked me for guidance on what all was needed for them to be able to enable audit logging for managed SSO instances so that they could create a ticket for it.  I pointed them to the existing ticket and told them I would check into why this PR merge was still pending.

@abstractj It looks like the PR merge may have stalled out waiting for your review despite it having the  required approving review from @slaskawi.  Is there something else needed here before we can unblock the merge?

pb82:
@stianst Makes sense, I can change the default to disabled. Is it ok to have those options as part of the Realm CR?

stianst:
As the Realm CR should just reflect the Realm JSON file [1] it should for sure be possible to set all these options.

I would argue that they should ideally not be listed in the Realm CR unless explicitly set by the user. If we somehow start adding defaults in the Realm CR that may easily not be consistent with defaults set by the Keycloak server. In the future we also want to improve the import/export of Realms so default values are not included in the json file.

[1] https://github.com/keycloak/keycloak/blob/master/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java#L153

pb82:
Logging is now optional and if not provided, they will not be enabled and not show up in the CR. They can be passed to the realm by adding this under spec:

```yaml
spec:
    ...
    eventsEnabled: true
    adminEventsEnabled: true
    adminEventsDetailsEnabled: true
```

slaskawi:
@stianst @abstractj This one is ready to be merged.

slaskawi:
@stianst We probably want this in `8.0.2`

PR #126, opened 2020-01-27
add generated moqs

briangallagher:
## Additional Information
The generated moqs are required for testing the Integreatly Operator

## Verification Steps
Ensure all tests are passing


briangallagher:
@slaskawi Can you review please

grdryn:
@slaskawi these are automatically generated by running `go generate ./...` which picks up on lines such as this: https://github.com/keycloak/keycloak-operator/blob/master/pkg/common/client.go#L727

Why would such `go:generate` directive comments be there if you didn't want to generate these moq files?

It would be really convenient for them to be here, so that people who want a mock implementation don't have to hand-create it (and manually keep it up-to-date if the interface changes) :slightly_smiling_face:

philbrookes:
I agree with Ger here, it does make the most sense to keep the mocks in this repo, not only does that mean that the mock would be automatically kept up to date, but it also means that anyone using your code can test it with the mocks that come with it.

The alternative is them having to create a mock and then manually keep it up to date, which is duplicating the mock in all the consuming repos.

abstractj:
@grdryn @philbrookes speaking as someone from outside, because you guys have way more experience than me on Operators, I do appreciate the fact that's something convenient for developers, I really do. 

At the same time, introducing something that we don't have any plan to use introduces a maintenance cost. The reason why that line of code is still there, @slaskawi can correct if I'm wrong, it's because we forgot to remove.

If you guys reach a consensus that using mocks it's the best thing to do I'm aboard with that. But looking at other Operators on GitHub I couldn't find any justification to have it.

slaskawi:
Yes, as @abstractj mentioned, this mock generation instruction is something we forgot to delete when moving the code from the old Operator to this one. 

Just to add a bit of a context - when initially porting the Keycloak Client, we created a [KEYCLOAK-12025](https://issues.redhat.com/browse/KEYCLOAK-12025) ticket to extract it into a separate repo. I still think it would make sense. Then, having mocks there would make perfect sense. Perhaps, if you have time - you could split it out?

stianst:
For now we will close this PR, based on comments from @slaskawi and @abstractj. However, if you are still interested in this change please open a discussion on our dev mailing list (see keycloak.org/community) and we can discuss there if this is something we should include or not.

PR #128, opened 2020-01-30
KEYCLOAK-12841: allow creating users without credentials

pb82:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-12841

## Additional Information
If no credentials are provided when creating users, the operator should not assign random ones. There are legitimate use cases where users without credentials are needed.

## Verification Steps

1. Deploy Keycloak from this branch
2. Create a realm
3. Create a user in the realm
4. Check the user output secret: it should only contain a username but no password
5. Login to the admin console and check the user: in the credential section there should be no 'Disable credentials' section

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

## Additional Notes

Note that once credentials are assigned to a user, it is not possible to remove them via the CR. Keycloak ignores an update with removed credentials.

### DIFF
--- a/pkg/common/controller_utils.go
+++ b/pkg/common/controller_utils.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/keycloak/keycloak-operator/pkg/model"
-
 	"github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


stianst:
From my perspective this is a correct fix. The operator should not apply different defaults to creating users (or anything else) than what Keycloak does. As such no passwords should be generated if credentials are not included.

stianst:
Question - does this need any updates to documentation? Does the documentation state something around passwords being generated?

pb82:
@stianst The documentation inside the operator does not say anything about generated passwords. It only gives an example that includes credentials.

@slaskawi Is there another location where you keep operator docs?

slaskawi:
@pb82 At the moment - no. But once we're unblocked, I'll provide proper updates to `keycloak-documentation` repo.

slaskawi:
@stianst @abstractj This one is ready to be merged.

PR #133, opened 2020-02-04
Adding delete permission to the endpoint apiGroup

nbonavia:
Adding delte permission to the endpoint apiGroup so that Keycloak Operator does not fail when setting ownerRef while creating an external database endpoint

## Additional Information
When enabling external database, the Operator fails with the following error message: -
```
endpoints “keycloak-postgresql” is forbidden: cannot set an ownerRef on a
    resource you can’t delete: , <nil>
```

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -21,6 +21,7 @@ rules:
   - create
   - update
   - watch
+  - delete
 - apiGroups:
   - apps
   resources:


slaskawi:
@stianst We probably want this in `8.0.2`

PR #125, opened 2020-01-24
fix an few issues discovered when test monitoring

wei-lee:
## JIRA ID
https://issues.redhat.com/browse/INTLY-5027

## Verification Steps

Check the alerts here: https://prometheus-route-rhmi-middleware-monitoring.apps.weili-43.w3g9.s1.devshift.org/alerts. You should see alerts for 2 instances of keycloak are working

## Checklist:
- [ ] Verified by team member
- [ ] Comments where necessary
- [ ] Automated Tests
- [ ] Documentation changes if necessary

PR #135, opened 2020-02-06
Keycloak 12917: Fix PodMonitor Naming to prevent Prometheus-Operator from breaking

roberth1988:
Hi Guys,

me again. Once I had fixed the rights for the operator [KEYCLOAK-12915] (so it was able to work with prometheus-operator) it broke the config of the prometheus operator.

Background is that it creates currently a PodMonitor as well a ServiceMonitor with the *ApplicationName* which result into a duplicate scrape job within the config since they have both the same name.

{code}
 level=error ts=2020-02-01T00:19:35.770Z caller=main.go:587 msg="Error reloading config" err="couldn't load configuration (--config.file=\"/etc/prometheus/config_out/prometheus.env.yaml\"): parsing YAML file /etc/prometheus/config_out/prometheus.env.yaml: found multiple scrape configs with job name \"keycloak/keycloak/0\""
{code}

I would suggest adding a (in future configurable?) postfix on the PodMonitor. Which I did here in the PR. This solves the issue with the duplicate scrape job name.

### DIFF
--- a/pkg/model/pod_monitor.go
+++ b/pkg/model/pod_monitor.go
@@ -12,7 +12,7 @@ import (
 func PodMonitor(cr *v1alpha1.Keycloak) *monitoringv1.PodMonitor {
 	return &monitoringv1.PodMonitor{
 		ObjectMeta: v12.ObjectMeta{
-			Name:      ApplicationName,
+			Name:      ApplicationName + "-pod",
 			Namespace: cr.Namespace,
 			Labels: map[string]string{
 				"monitoring-key": MonitoringKey,


slaskawi:
@davidffrench @pb82 @wei-lee Would you mind to have a look at it?

From my perspective - it looks good.

wei-lee:
@slaskawi This issue is already fixed as part of this PR: https://github.com/keycloak/keycloak-operator/pull/125. We just need to merge it.

cc @davidffrench @pb82

slaskawi:
@roberth1988 Could you please check if solution proposed by @wei-lee (#125) for for you?

roberth1988:
@slaskawi works for me. I will close that PR therefore. Thanks guys!

PR #140, opened 2020-02-12
KEYCLOAK-12787: Fix external Database Support

roberth1988:
This fixes that the Database name wasn't taken from the Secret which resulted in a non working setup.


pb82:
A similar fix is included in https://github.com/keycloak/keycloak-operator/pull/137

slaskawi:
@roberth1988 Could you please check if #137 solves your problem. If it is, please close this Pull Request.

slaskawi:
Let me close this Pull Request, @roberth1988. I believe it has been addressed by @pb82 's Pull Request.

PR #142, opened 2020-02-14
KEYCLOAK-13008: Set the ClientSecret to be written as CLIENT_SECRET

roberth1988:
Hi guys,

whenever a KeycloakClient is added with:

```
clientAuthenticatorType: client-secret
```

a secret is created with 

```
CLIENT_ID
CLIENT_SECRET
```

But unfortunately the CLIENT_SECRET is always empty and therefore can't be used the way its meant.

Thats fixed within this PR.

Offtopic: Its a pain to create JIRA Ticket ... then Github-Issue .. mark all things with the JIRA-Ticket...

To check it you can adjust:
```
---
  kind: KeycloakClient
  metadata:
    finalizers:
    - client.cleanup
    generation: 2
    name: example-name
    namespace: example-ns-name
  spec:
    client:
      bearerOnly: false
      clientAuthenticatorType: client-secret
      clientId: pay
      defaultClientScopes:
      - some-scopeshere1
      - some-scopeshere2
      - some-scopeshere3
      description: The nice description
      id: exampl-example
      protocol: openid-connect
      publicClient: false
      rootUrl: ${authBaseUrl}
      serviceAccountsEnabled: true
    realmSelector:
      matchLabels:
        iam.example.cloud/keycloak-app: example-iam
```

### DIFF
--- a/pkg/common/client_state.go
+++ b/pkg/common/client_state.go
@@ -35,6 +35,12 @@ func (i *ClientState) Read(context context.Context, cr *kc.KeycloakClient, realm
 		return nil
 	}
 
+	clientSecret, err := realmClient.GetClientSecret(cr.Spec.Client.ID, i.Realm.Spec.Realm.Realm)
+	if err != nil {
+		return err
+	}
+	cr.Spec.Client.Secret = clientSecret
+
 	err = i.readClientSecret(context, cr, i.Client, controllerClient)
 	if err != nil {
 		return err


slaskawi:
Thanks for the Pull Request @roberth1988 !

As I commented in [KEYCLOAK-13008](https://issues.redhat.com/browse/KEYCLOAK-13008), this seems like a wrong approach to solve this particular case.

However, you do have right - not all `Clients` requires `Secrets`. Some of them might be public (and therefore they don't use Secrets).

Could you please repurpose your Pull Request to solve this case as well? You're very close, the `getUpdatedClientSecretState` function needs to be changes as well and tread a `nil` Secret as something expected. Also, we would need another JIRA for this one.

roberth1988:
@slaskawi I will take care of it on the weekend.

stianst:
@slaskawi see my comment in the JIRA. This looks indeed to be an issue in the Operator, not Keycloak server.

slaskawi:
After a short discussion on [KEYCLOAK-13008](https://issues.redhat.com/browse/KEYCLOAK-13008), this PR looks good to me. Thanks for the contribution @roberth1988 !

slaskawi:
@stianst May I ask you to merge this one?

PR #148, opened 2020-02-24
Keycloak 13065 operator sdk upgrade

slaskawi:
## JIRA ID

[KEYCLOAK-13065](https://issues.redhat.com/browse/KEYCLOAK-13065)

## Additional Information

This Pull Request provides an update to the latest Operator SDK version

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


slaskawi:
@davidffrench @pb82 May I ask you for a review?

pb82:
@slaskawi e2e tests are failing with `no matches for kind \"Keycloak\" in version \"keycloak.org/v1alpha1`. This seems to be caused by an invalid Keycloak CRD:

```
The CustomResourceDefinition "keycloaks.keycloak.org" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[instances].default: Forbidden: must not be set
```

slaskawi:
Thanks for checking this @pb82 !

It seems the defaults will be supported starting from K8s `1.17` (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting).

I probably should remove them altogether...

go.mod
```diff
@@ -1,65 +1,52 @@
...
 	k8s.io/api v0.0.0
 	k8s.io/apimachinery v0.0.0
 	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d
-	sigs.k8s.io/controller-runtime v0.3.0
+	sigs.k8s.io/controller-runtime v0.4.0
+
 )
 
 // Pinned to kubernetes-1.15.4
```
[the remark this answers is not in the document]
slaskawi: Good point! Thanks for spotting this!

slaskawi:
> One additional change that should be made is chaning the make gen/code command to use operator-sdk generate crds instead of open-api

@davidffrench I have also been thinking about this... The `opem-api` has been extracted to a separate command. Let me see if it's stable enough to be included (or maybe we need to wait a few releases more to do it).

slaskawi:
@davidffrench Makefile updated.

Makefile
```diff
@@ -108,7 +108,10 @@ code/compile:
 .PHONY: code/gen
 code/gen:
 	operator-sdk generate k8s
-	operator-sdk generate openapi
+	operator-sdk generate crds
+	# This is a copy-paste part of `operator-sdk generate openapi` command (suggested by the manual)
```
davidffrench: @slaskawi Where are you seeing that it is suggested by the manual? It was deprecated in favour of `generate crds` in operator-sdk version `0.13`[1], it does say to use it directly for OpenAPI code generation but I don't believe that is needed in this operator.

[1] https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md#notable-changes

slaskawi: Here's my output of executing `$ operator-sdk generate openapi` command:

```
[Deprecation notice] The 'operator-sdk generate openapi' command is deprecated!

 - To generate CRDs, use 'operator-sdk generate crds'.
 - To generate Go OpenAPI code, use 'openapi-gen'. For example:

      # Build the latest openapi-gen from source
      which ./bin/openapi-gen > /dev/null || go build -o ./bin/openapi-gen k8s.io/kube-openapi/cmd/openapi-gen

      # Run openapi-gen for each of your API group/version packages
      ./bin/openapi-gen --logtostderr=true -o "" -i ./pkg/apis/<group>/<version> -O zz_generated.openapi -p ./pkg/apis/<group>/<version> -h ./hack/boilerplate.go.txt -r "-"

INFO[0000] Running OpenAPI code-generation for Custom Resource group versions: [keycloak:[v1alpha1], ] 
INFO[0008] Code-generation complete.                    
INFO[0008] Running CRD generator.                       
INFO[0008] CRD generation complete.
```

So it seems replacing `operator-sdk generate openapi` requires using `openapi-gen` command.

davidffrench: I see it here after looking - https://github.com/operator-framework/operator-sdk/blob/76b9041029e3d3d19d95ee32afebacdd8999c37a/doc/sdk-cli-reference.md#openapi

I think but maybe wrong that since all the validation is now done through the CRD itself, the `zz_generated.openapi.go` file is no longer needed but this is just a hunch. I would be curious with a brand new operator, if you added an api and ran the generation, would it generate this file at all.

slaskawi: From my understanding, the Open API stuff is needed to expose and document API Server REST endpoints for our CRDs. If my train of thought is right, the `zz_generated.openapi.go` would still be needed. But I guess we need to ask some OpenShift folks to clarify this.

As for now, I would like to preserve this file. Maybe we'll kill it in subsequent release but I would like to double check if we don't break anything.

davidffrench: Seems reasonable @slaskawi .

slaskawi:
@stianst May I ask you to merge this one?

PR #149, opened 2020-02-25
[KEYCLOAK-13090] Add support for adding extra Environment Variables

roberth1988:
Hi Guys,

the follow up PR will contain changes to be able to define custom Environment-Variables within the CRD.

For example sometimes you need to pass these things:

```
...
- name: PROXY_ADDRESS_FORWARDING
   value: "true"
- name: WILDFLY_LOGLEVEL
   value: DEBUG
- name: KEYCLOAK_LOGLEVEL
   value: DEBUG
- name: JDBC_PARAMS
   value: "?prepareThreshold=0&autoCommit=false"
...
```

 and even more which is not possible yet.

### DIFF
--- a/pkg/model/keycloak_deployment.go
+++ b/pkg/model/keycloak_deployment.go
@@ -27,7 +27,7 @@ func GetServiceEnvVar(suffix string) string {
 }
 
 func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
-	return []v1.EnvVar{
+	defaultEnv := []v1.EnvVar{
 		// Database settings
 		{
 			Name:  "DB_VENDOR",
@@ -124,6 +124,17 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 			Value: "/var/run/secrets/kubernetes.io/serviceaccount/*.crt",
 		},
 	}
+
+	if len(cr.Spec.ExtraEnv) > 0 {
+		for k, v := range cr.Spec.ExtraEnv {
+			defaultEnv = append(defaultEnv, v1.EnvVar{
+				Name: k,
+				Value: v,
+			})
+		}
+	}
+
+	return defaultEnv
 }
 
 func KeycloakDeployment(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) *v13.StatefulSet {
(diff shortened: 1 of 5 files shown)


pkg/model/keycloak_deployment.go
```diff
@@ -124,6 +124,17 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 			Value: "/var/run/secrets/kubernetes.io/serviceaccount/*.crt",
 		},
 	}
+
+	if len(cr.Spec.ExtraEnv) > 0 {
+		for k, v := range cr.Spec.ExtraEnv {
+			defaultEnv = append(defaultEnv, v1.EnvVar{
+				Name: k,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				Name:  k,
```

slaskawi:
Having in mind recent discussion in [KEYCLOAK-12409](https://issues.redhat.com/browse/KEYCLOAK-12409), I think the argumentation for rejecting this Pull Request is similar. Supporting things like `PROXY_ADDRESS_FORWARDING` or using custom JDBC Connection URL should be implemented as core functionality of the Operator. Modifying Environment Variables to do that is not the right approach.

@stianst If you agree with me, please close this Pull Request.

stianst:
+1 Operator should have a more opinionated approach, and any configuration provided should be very simple and high-level. There's also the fact that the Operator under the covers right now uses Keycloak images as well as RH-SSO images, where environment variables in the two are not consistent, so you need an "abstraction". In the future we will also introduce support for a Keycloak.X which will not have the same environment variables. The Operator should not have "use at your own risk" type of configuration options.

andrew-waters:
@slaskawi @stianst is there a tracking ticket to enable this in the operator? We're facing issues with a new install (and new experience) with Keycloak.

The issue we have is that we use an external IngressRoute (Traefik 2.0) which selects the `keycloak-discovery` service created by the operator. That allows us to load the dashboard but connections to the `/auth` endpoint result in JS errors:

```
Mixed Content: The page at 'https://<domain>/auth/admin/master/console/' was loaded over HTTPS, but requested an insecure script 'http://<domain>/auth/js/keycloak.js?version=me0et'. This request has been blocked; the content must be served over HTTPS.
```

The way around this as I read it is to set `PROXY_ADDRESS_FORWARDING` to true which this PR allowed according to the description. Can you advise on what the best approach is (happy to move to another issue if preferable)

PR #150, opened 2020-03-06
KEYCLOAK-13018: Ensure state contains the correct user

JameelB:
The `FindUserByUsername` function returns the first user which contains the string specified in the search query. Since this function is used for managing the state of a KeycloakUser, if a KeycloakUser was created or deleted with a username that is a substring of a user that already exists in the realm and is the first result of that query, it will update or delete that user instead.

Example: 
1. `test-admin` already exists in the realm
2. A new KeycloakUser custom resource was created with the username `test`
3. This will update `test-admin` with the `test` user's details.
4. The `test` KeycloakUser custom resource status is set to reconciled.

The following changes ensures that the user found by this function returns the user that matches the correct username. The parameter to the request `search=%s` was also changed to `username=%s` to filter users by username instead of all the fields.

## JIRA ID
[KEYCLOAK-13018](https://issues.redhat.com/browse/KEYCLOAK-13018)
This is also tracked in the Integreatly board: [INTLY-5813](https://issues.redhat.com/browse/INTLY-5813)

## Additional Information
Steps to reproduce this bug in RHMI can be found in [INTLY-5813](https://issues.redhat.com/browse/INTLY-5813)

## Verification Steps
1. Run the keycloak operator locally
2. Create users in a realm
3. Create a user by creating a KeycloakUser CR that has a username which is a substring of an existing user in the realm.
4. Ensure that the user is created successfully and that the existing user(s) was not updated.
5. Ensure the KeycloakUser CR of the newly created user has the correct id
6. Delete the KeycloakUser CR created in step 3.
7. Ensure that the user was deleted successfully
8. Ensure no other users were deleted from the realm.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/common/client.go
+++ b/pkg/common/client.go
@@ -38,11 +38,11 @@ type Client struct {
 type T interface{}
 
 // Generic create function for creating new Keycloak resources
-func (c *Client) create(obj T, resourcePath, resourceName string) error {
+func (c *Client) create(obj T, resourcePath, resourceName string) (string, error) {
 	jsonValue, err := json.Marshal(obj)
 	if err != nil {
 		logrus.Errorf("error %+v marshalling object", err)
-		return nil
+		return "", nil
 	}
 
 	req, err := http.NewRequest(
@@ -52,7 +52,7 @@ func (c *Client) create(obj T, resourcePath, resourceName string) error {
 	)
 	if err != nil {
 		logrus.Errorf("error creating POST %s request %+v", resourceName, err)
-		return errors.Wrapf(err, "error creating POST %s request", resourceName)
+		return "", errors.Wrapf(err, "error creating POST %s request", resourceName)
 	}
 
 	req.Header.Set("Content-Type", "application/json")
@@ -61,12 +61,12 @@ func (c *Client) create(obj T, resourcePath, resourceName string) error {
 
 	if err != nil {
 		logrus.Errorf("error on request %+v", err)
-		return errors.Wrapf(err, "error performing POST %s request", resourceName)
+		return "", errors.Wrapf(err, "error performing POST %s request", resourceName)
 	}
 	defer res.Body.Close()
 
 	if res.StatusCode != 201 && res.StatusCode != 204 {
-		return fmt.Errorf("failed to create %s: (%d) %s", resourceName, res.StatusCode, res.Status)
+		return "", fmt.Errorf("failed to create %s: (%d) %s", resourceName, res.StatusCode, res.Status)
 	}
 
 	if resourceName == "client" {
@@ -74,22 +74,24 @@ func (c *Client) create(obj T, resourcePath, resourceName string) error {
 		fmt.Println("user response ", string(d))
 	}
 
-	return nil
+	location := strings.Split(res.Header.Get("Location"), "/")
+	uid := location[len(location)-1]
+	return uid, nil
 }
 
-func (c *Client) CreateRealm(realm *v1alpha1.KeycloakRealm) error {
+func (c *Client) CreateRealm(realm *v1alpha1.KeycloakRealm) (string, error) {
 	return c.create(realm.Spec.Realm, "realms", "realm")
 }
 
-func (c *Client) CreateClient(client *v1alpha1.KeycloakAPIClient, realmName string) error {
+func (c *Client) CreateClient(client *v1alpha1.KeycloakAPIClient, realmName string) (string, error) {
 	return c.create(client, fmt.Sprintf("realms/%s/clients", realmName), "client")
 }
 
-func (c *Client) CreateUser(user *v1alpha1.KeycloakAPIUser, realmName string) error {
+func (c *Client) CreateUser(user *v1alpha1.KeycloakAPIUser, realmName string) (string, error) {
 	return c.create(user, fmt.Sprintf("realms/%s/users", realmName), "user")
 }
 
-func (c *Client) CreateFederatedIdentity(fid v1alpha1.FederatedIdentity, userID string, realmName string) error {
+func (c *Client) CreateFederatedIdentity(fid v1alpha1.FederatedIdentity, userID string, realmName string) (string, error) {
 	return c.create(fid, fmt.Sprintf("realms/%s/users/%s/federated-identity/%s", realmName, userID, fid.IdentityProvider), "federated-identity")
 }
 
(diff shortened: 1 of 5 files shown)


pkg/common/client.go
```diff
@@ -179,15 +179,18 @@ func (c *Client) FindUserByEmail(email, realm string) (*v1alpha1.KeycloakAPIUser
 }
 
 func (c *Client) FindUserByUsername(name, realm string) (*v1alpha1.KeycloakAPIUser, error) {
-	result, err := c.get(fmt.Sprintf("realms/%s/users?first=0&max=1&search=%s", realm, name), "user", func(body []byte) (T, error) {
+	result, err := c.get(fmt.Sprintf("realms/%s/users?username=%s", realm, name), "user", func(body []byte) (T, error) {
```
matskiv: This API endpoint has a max field with a default of 100.
Docs: https://www.keycloak.org/docs-api/9.0/rest-api/index.html#_getusers

We should pass it to get all results. Maybe `0` value will work.

slaskawi: In Keycloak there's a unique constraint on username (see [this piece of the code](https://github.com/keycloak/keycloak/blob/b478472b3578b8980d7b5f1642e91e75d1e78d16/model/jpa/src/main/resources/META-INF/jpa-changelog-2.5.0.xml#L82)). So gathering all users doesn't make sense to me as you will either get 1 entity back or nothing.

So I would prefer to put `first=0&max=1` back or use `search` instead of `username`.

stianst:
Would it not be better if the Operator added the ID of the user to the User CR after it has been created? That way the Operator knows it is managing the correct user, and not just some user that happens to have the same username. 

Bear in mind that username can be mutable, so in theory someone could change the username, which then allows someone else to use the previously used username, resulting in the Operator "managing" the wrong user.

pkg/common/client.go
JameelB: @slaskawi Even if the username has a unique constraint, it doesn't stop it from being a substring of another username/containing another username in it. 

The problem with using `max=1` is that the API `username` or `search` filter seems to return all users that contains the specified string. The first result of this request isn't necessarily the correct user even if the username is unique.

I would agree with @stianst suggestion of adding the uid to the User CR. This way we're ensuring that the correct user is always returned especially if the username value can be changed.

matskiv: @slaskawi From the observed behavior it seemed like the API endpoint we use here^ is using this query:
https://github.com/keycloak/keycloak/blob/93a1374558643ee31e717c1dfacecb8ab8d95f68/model/jpa/src/main/java/org/keycloak/models/jpa/entities/UserEntity.java#L47-L48
and the query uses `like` instead of exact matching.

JameelB: @matskiv Looks like `max=0` doesn't disable this but rather just sets the maximum result to 0, being an empty list. I've set it to `max=-1` instead.

JameelB:
@stianst I'm trying to figure out the best way of getting the user `ID` back to the KeycloakUser CR. 
The create user endpoint doesn't seem to return any info on the user it has created. I also tried passing in an id as a field to the request but Keycloak doesn't seem to take it into account and generates an id for the new user anyway. 

We could call `FindUserByUsername` directly after creating a user which should give us the user id. By this time, no one should have had a chance to change the username yet. We can then add that onto the KeycloakUser CR status and use this ID to get the User in future requests. wdyt? 

However, it would be great to return the user id of a created user in the response. This would make it easier to put this info back to the KeycloakUser CR. Would it be possible to add this change to the API?

slaskawi:
@JameelB When you create a user, you should get a User ID Back from Keycloak. See this piece of the code: https://github.com/keycloak/keycloak/blob/master/services/src/main/java/org/keycloak/services/resources/admin/UsersResource.java#L138

stianst:
It is returned in the Location header - kinda awkward, but it's there ;)

JameelB:
Thanks @slaskawi @stianst I see it there now :smile: I've just been looking at the response body all this time

pkg/common/client_test.go
```diff
@@ -100,9 +123,12 @@ func TestClient_CreateUser(t *testing.T) {
 	// given
 	user := getDummyUser()
 	realm := getDummyRealm()
+	dummyUserId := "dummy-user-id"
```
golangcibot: var `dummyUserId` should be `dummyUserID` (from `golint`)

pkg/common/client_test.go
```diff
@@ -100,9 +123,12 @@ func TestClient_CreateUser(t *testing.T) {
...
 
 	handler := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		assert.Equal(t, fmt.Sprintf(UserCreatePath, realm.Spec.Realm.Realm), req.URL.Path)
+		locationUrl := fmt.Sprintf("http://dummy-keycloak-host/%s", UserGetPath)
```
golangcibot: var `locationUrl` should be `locationURL` (from `golint`)

JameelB:
@stianst @slaskawi I've update this PR with the requested changes.

- The generic create function for the Keycloak client now returns the resource id.
- This id should now be added to the KeycloakUser custom resource after creation.
- Updated the function which reads the user state to get the user by ID `GetUser` instead of finding the user by username `FindUserByUsername`.
- I've left the changes I've made to `FindUserByUsername` since I believe this change is still valid.

JameelB:
@stianst @slaskawi Can we get a follow up on this when you get the time?

pkg/common/client_test.go
```diff
@@ -115,11 +141,13 @@ func TestClient_CreateUser(t *testing.T) {
 	}
 
 	// when
-	err := client.CreateUser(user, realm.Spec.Realm.Realm)
+	uid, err := client.CreateUser(user, realm.Spec.Realm.Realm)
 
 	// then
 	// correct path expected on httptest server
 	assert.NoError(t, err)
+
```
slaskawi: Please remove this empty line.

pb82:
:eyes:

pkg/common/cluster_actions.go
```diff
@@ -114,14 +114,22 @@ func (i *ClusterActionRunner) CreateRealm(obj *v1alpha1.KeycloakRealm) error {
 	if i.keycloakClient == nil {
 		return errors.New("cannot perform realm create when client is nil")
 	}
-	return i.keycloakClient.CreateRealm(obj)
+
+	if _, err := i.keycloakClient.CreateRealm(obj); err != nil {
```
pb82: We don't really need the if here, could be just:
```
_, err := i.keycloakClient.CreateRealm(obj);
return err
```

JameelB:
@pb82 @slaskawi I've updated this PR with the requested changes. I've also squashed the commit with the JIRA prefix on it.

pkg/common/cluster_actions.go
```diff
@@ -187,7 +198,10 @@ func (i *ClusterActionRunner) AssignRealmRole(obj *v1alpha1.KeycloakUserRole, us
 	if i.keycloakClient == nil {
 		return errors.New("cannot perform role assign when client is nil")
 	}
-	return i.keycloakClient.CreateUserRealmRole(obj, realm, userID)
+
+	_, err := i.keycloakClient.CreateUserRealmRole(obj, realm, userID)
+	return err
+
```
golangcibot: unnecessary trailing newline (from `whitespace`)

pb82:
@slaskawi Looks good from my side. Verified locally. Creating users works as expected and the ID is added to the spec.

PR #152, opened 2020-03-10
KEYCLOAK-13159 Add `patch` verb to "" apiGroup

sergioifg94:
The `k8s/client-go` library logic attempts to update an event instead of creating it when it already
exists. However, the operator doesn't have permissions to patch events, so it will fail and
show a "Server rejected event" error log.

Add the `patch` verb to the `""` apiGroup in `deploy/role.yaml` in order to allow the operator
to perform the update.

## JIRA ID
[KEYCLOAK-13159](https://issues.redhat.com/browse/KEYCLOAK-13159)
Also tracked in the Integreatly board [INTLY-5812](https://issues.redhat.com/browse/INTLY-5812)

## Additional Information
* The logic in `k8s/client-go` for deciding when to patch events is based on the `Event.Count` field [1]
* The `operator-sdk` CLI already includes the `patch` verb in the scaffolding for the `role.yaml` file [2]

[1. `k8s/client-go` GitHub repository](https://github.com/kubernetes/client-go/blob/06eb1244587a17d4994ff9de4ce93756a2f16f0b/tools/record/event.go#L189)
[2. `operator-sdk` GitHub repository](https://github.com/operator-framework/operator-sdk/blob/master/internal/scaffold/role.go#L188) 

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/deploy/cluster_roles/cluster_role.yaml
+++ b/deploy/cluster_roles/cluster_role.yaml
@@ -18,6 +18,7 @@ rules:
       - list
       - get
       - create
+      - patch
       - update
       - watch
       - delete
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -19,6 +19,7 @@ rules:
   - list
   - get
   - create
+  - patch
   - update
   - watch
   - delete


slaskawi:
@davidffrench @pb82 This one seems like an incorrect logic implemented somewhere? Have you seen this behavior before?

pb82:
@slaskawi The operator is generating events in error cases (e.g. https://github.com/keycloak/keycloak-operator/blob/master/pkg/controller/keycloak/keycloak_controller.go#L203). Kubernetes does need to update events, for example to group multiple events into a single one if more of the same type are firing. It looks like it's using patch instead of update (patch allows you to only submit a partial object and kubernetes applies the partial changes). So i would say its legitimate.

slaskawi:
Thanks a lot, @pb82 !

@sergioifg94 May I also ask you to update the cluster roles. Once it's done, I'll merge it.

sergioifg94:
@slaskawi Perfect, will do

sergioifg94:
@slaskawi Just pushed that change

PR #123, opened 2020-01-18
Add deployment kustomization

mhutter:
## JIRA ID
_none_

## Additional Information
This PR allows the keycloak-operator to be installed using [kustomize](https://kustomize.io/). Example:

```yaml
---
namespace: keycloak-operator
commonLabels:
  app.kubernetes.io/name: keycloak-operator
  app.kubernetes.io/managed-by: flux
bases:
  - github.com/keycloak/keycloak-operator//deploy
```

## Verification Steps
Create a `kustomize.yaml`:

```yaml
---
namespace: keycloak-operator
commonLabels:
  app.kubernetes.io/name: keycloak-operator
  app.kubernetes.io/managed-by: flux
bases:
  - github.com/mhutter/keycloak-operator//deploy
```

    kubectl apply -k . --dry-run

## Checklist:
- [ ] Verified by team member
- [ ] Comments where necessary
- [ ] Automated Tests
- [ ] Documentation changes if necessary

PR #166, opened 2020-04-14
Adds kustomize.

pierreozoux:
We find it easier to deploy.

We use it like this:
https://git.indie.host/liiibre/infrastructure/operators/-/blob/master/keycloack/kustomization.yaml

Thanks :)

### DIFF
--- a/deploy/kustomization.yaml
+++ b/deploy/kustomization.yaml
@@ -1,0 +1,10 @@
+resources:
+- ./crds/keycloak.org_keycloakbackups_crd.yaml
+- ./crds/keycloak.org_keycloakclients_crd.yaml
+- ./crds/keycloak.org_keycloakrealms_crd.yaml
+- ./crds/keycloak.org_keycloaks_crd.yaml
+- ./crds/keycloak.org_keycloakusers_crd.yaml
+- ./role.yaml
+- ./role_binding.yaml
+- ./service_account.yaml
+- ./operator.yaml


slaskawi:
Thank you for contribution, @pierreozoux !

Unfortunately we are not keen to merge as. Kustomize is something we don't plan to support. 

For the record - we already had a similar Pull Request (which also got rejected) here: https://github.com/keycloak/keycloak-operator/pull/123

pierreozoux:
Hum, this is sad :/ This is the only place where it can be hosted :/
I can take care of supporting it. How can I become a contributor to this project?

slaskawi:
Unfortunately, that's not an option. 

In parallel to maintaining this project, we're working on a long-term vision for Keycloak Operator ([link](https://github.com/keycloak/keycloak-community/pull/54)). We envision the Operator as an opinionated way of installing and managing Keycloak. Therefore, we can't accept additional deployment options, which we won't be able to maintain. 

I'm very sorry, but the suggested way of installing this Operator is by using OLM and Operatorhub.

pierreozoux:
I know what is OLM, been following this since almost the beginning I think.
I find it promising, and once it will be more stable I'll probably install it.
In the mean time, I use kustomize to manage my operators.
I find it a bit sad that you dont embrace the diversity of your users. I understand that you have a product to manage, it is not asking a new weird functionality, it is just some lines of yaml. And it is the 2nd PR you receive, it should be a signal that tell you where your users are.

Anyway :) I understand, but still a bit frustrated :)

Still, I really appreciate your work, and really looking forward this operator development :)

Wish you all the best :)

slaskawi:
>  I find it promising, and once it will be more stable I'll probably install it.

@pierreozoux Thanks again for the comment!

May I ask you how exactly do you use Kustomize and what issues did you have with the OLM? Perhaps there is something we could help or do to help you?

stianst:
As @slaskawi mentioned supporting alternative ways of doing things is a lot of extra effort on maintenance. Sure, it's a line or two of YAML added, but what about documentation and testing? That means documenting two different approaches to installing the Operator, as well as having to develop automated testing around installing/managing the Operator using Kustomize instead of OLM.

From what I know (and I'm no Operator expert by far) OLM is the recommended way to install and manage operators both on Kubernetes and OpenShift, so I'm not very keen on adding additional non-recommended ways here.

pierreozoux:
In Reddhat it is the recommanded way. In the  wild, people use:
 - kustomize (me and the other PR :) )
 - helm
 - ansible
 - bash (kubedb operator)
 - git clone kubectl apply 

I dont think that OLM is the main way. Not yet at least.
I'm fine if there is no doc and no tests actually,  unsupported features and I take care of coming here and making a PR of it is broken

Kampe:
The opinionated nature of OLM is why we removed it entirely from our clusters, its connection to catalog servers just isn't ideal.  

Everything can be installed via kustomization and strategic merge patching, simply by including a small bit of yaml in the repo, causing no issues downstream.

trevorloranger:
Hi @stianst, I came across this thread while searching for a helm chart for the Keycloak Operator. Based on the comments by you and others I can draw some reasonable conclusions but I will ask anyways:
- Is it accurate to say that there are no future plans for providing helm charts for the operator? Additionally, will the OLM be the only "officially" supported deployment mechanism for the operator for the foreseeable future?
- Would you be willing to accept a helm chart contribution provided there is proper documentation, testing, etc.? I say this because we're actively looking into this.

Thank you!

trevorloranger:
@slaskawi maybe you can comment on the above? Thank you!

sum2000:
@slaskawi is there a plan to add support for the helm chart? I have a PR ready and wondering if that's something you guys will be open to?

slaskawi:
Unfortunately no. We added kustomize only because it's part of the Operator
Framework 1.0.

PR #168, opened 2020-04-17
Lower alert severity for memory based alerts

david-martin:
## JIRA ID
https://issues.redhat.com/browse/INTLY-6614

## Additional Information
Some discussion in https://issues.redhat.com/browse/INTLY-6614, the gist of which is a review of the critical alerts and if they should be critical (i.e. they page someone when they fire).

My thinking is that only problems that affect a potential service level objective (SLO) should be critical.
In this case, its:
- `KeycloakInstanceNotAvailable` if the service isn't available
- `KeycloakAPIRequestDuration90PercThresholdExceeded` & `KeycloakAPIRequestDuration99.5PercThresholdExceeded` if the service isn't responding as fast as expected

All other alerts would help with debugging & point at potential causes as to why one of the critical alerts are firing, but as long as the SLO is being met, no-one needs to be paged.

## Verification Steps
1. Install operator in a cluster with prometheus-operator
2. Verify the alerts are created with the expected severity

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


david-martin:
ping @pb82 @slaskawi

slaskawi:
Thanks @david-martin !

david-martin:
Wow, thanks for the quick turnaround @slaskawi 👍

PR #172, opened 2020-04-20
KEYCLOAK-13974 Dockerfile cleanup

slaskawi:
[KEYCLOAK-13974](https://issues.redhat.com/browse/KEYCLOAK-13974)

A long while ago, we discussed permissions issue in our Keycloak Container ([link1](https://github.com/keycloak/keycloak-containers/pull/259/files)[link2](https://groups.google.com/forum/#!topic/keycloak-dev/JWUErpYrZAA)). The outcome of the discussion is that a binary just needs to be owned by a `root` group and we don't really need any extra user. 

I'm not sure if this is the best moment, but perhaps we should do a permission clean up in our Dockerfile here as well? WDYT @stianst @abstractj @douglaspalmer @iankko @davidffrench ? If you think that's a good idea, I'll convert it into a proper Pull Request.

The image might be tested using: `slaskawi/keycloak-operator:cleanup`


iankko:
@slaskawi 

> This is just a Preview PR to check if we want this change or not.
> 
> A long while ago, we discussed permissions issue in our Keycloak Container ([link1](https://github.com/keycloak/keycloak-containers/pull/259/files)[link2](https://groups.google.com/forum/#!topic/keycloak-dev/JWUErpYrZAA)). The outcome of the discussion is that a binary just needs to be owned by a `root` group and we don't really need any extra user.
> 
> I'm not sure if this is the best moment, but perhaps we should do a permission clean up in our Dockerfile here as well? WDYT @stianst @abstractj @douglaspalmer @iankko @davidffrench ? If you think that's a good idea, I'll convert it into a proper Pull Request.

+1 on switching this preview PR into a real one (it's even fixing a CVE, see [this article](https://access.redhat.com/articles/4859371) for details).

Unsure if the Keycloak operator needs to run as privileged container already, or not (in order to be apply clusterwide settings). In any case weaking `/etc/passwd` permissions is discouraged in that article as dangerous, so better to avoid it IMHO

slaskawi:
Converting this into a "real" Pull Request.

slaskawi:
Just in case, here's the build: https://travis-ci.org/github/keycloak/keycloak-operator/builds/680493735

slaskawi:
@stianst @abstractj @hmlnarik @mhajas @iankko @pb82 @davidffrench May I ask for a review? This should be a quick one.

PR #173, opened 2020-04-21
Lower severity of rate based alerts

david-martin:
Discussed offline (& in https://issues.redhat.com/browse/INTLY-6614) 
I think it might be best to go down to just 1 critical alert where SRE need to investigate as a matter of urgency at this time i.e. the 'up' alert (KeycloakInstanceNotAvailable).

The response times alerts may become more important in time, but I don't think we know what immediate steps an SRE could take when paged, so i think lowering them to warning is reasonable.
There could be an immediate action to increase the number of instances (in the keycloak CR) while investigation continues, but i'd like to have that discussion with our SRE team first as there's resourcing concerns there. Also, they Keycloak CR gets reset back to 2 in RHMI whenever changed at the moment (David Ffrench is following up on that to allow changes)

TLDR: Just have the 'up' alert for at least 1 keycloak pod as critical.
All others as warning, but will add gained knowledge from discussions here to the SOP to help with debugging any warnings related to memory & GC.

## Additional Information
https://issues.redhat.com/browse/INTLY-6614

## Verification Steps
1. Install in a cluster with prometheus-operator watching the ns 
2. Verify the severity of alerts in Prometheus Alerts tab reflects the chagnes

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


david-martin:
ping @slaskawi for review

PR #174, opened 2020-04-21
KEYCLOAK-13853 Support access disabling

mhajas:
## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/controller/keycloak/keycloak_controller.go
+++ b/pkg/controller/keycloak/keycloak_controller.go
@@ -3,7 +3,13 @@ package keycloak
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/metadata"
+	config2 "sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	v1beta12 "k8s.io/api/policy/v1beta1"
 
@@ -157,6 +163,13 @@ func (r *ReconcileKeycloak) Reconcile(request reconcile.Request) (reconcile.Resu
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Keycloak")
 
+	// Do not Reconcile until StateManager is loaded
+	stateManager := common.GetStateManager()
+	_, keyExists := stateManager.GetState(common.RouteKind).(bool)
+	if !keyExists {
+		return reconcile.Result{Requeue: true}, nil
+	}
+
 	// Fetch the Keycloak instance
 	instance := &keycloakv1alpha1.Keycloak{}
 	err := r.client.Get(r.context, request.NamespacedName, instance)
@@ -174,6 +187,11 @@ func (r *ReconcileKeycloak) Reconcile(request reconcile.Request) (reconcile.Resu
 	// Read current state
 	currentState := common.NewClusterState()
 	err = currentState.Read(r.context, instance, r.client)
+	if err != nil {
+		return r.ManageError(instance, err)
+	}
+
+	err = r.removeOrphanedResourcesWithOwnerReference(instance)
 	if err != nil {
 		return r.ManageError(instance, err)
 	}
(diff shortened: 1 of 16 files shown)


pkg/controller/keycloak/keycloak_reconciler.go
```diff
@@ -66,6 +66,12 @@ func (i *KeycloakReconciler) reconcileExternalDatabase(desired *common.DesiredCl
 
 func (i *KeycloakReconciler) reconcileExternalAccess(desired *common.DesiredClusterState, clusterState *common.ClusterState, cr *kc.Keycloak) {
 	if !cr.Spec.ExternalAccess.Enabled {
+		if clusterState.KeycloakRoute != nil {
```
slaskawi, line 69: I totally agree, we should have something like this in the Operator. However, I think the implementation should be different.

My preference would be to inspect objects with owning reference pointing to the Operator (all objects that were created by this Operator has owning reference pointing to it). If any of those objects is not present in the Secondary Resources (from the status field of the CR), they should be removed. 

This way we would have a generic mechanism for removing objects that are no longer necessary. 

@pb82 @davidffrench @mhajas WDYT?

davidffrench: That would be good to have @slaskawi . I don't believe it will fix this issue as the route would exist in the secondary resources in the status block. The generic solution would work for outside sources adding the operator as an owner reference. Although I don't know if i have a problem with that either.

This code will fix an issue where you are switching the CR.spec.externalaccess.enabled from true to false.

slaskawi:
Hey @mhajas, Could you please tell me if you have time to work on this? Maybe we should close it for now and wait till it's ready?

mhajas:
@slaskawi I will find some time to work on this. This is not in the ready state because I am not sure I am on the correct path here. I would like someone to look at it and give me some feedback if it makes some sense.

pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -263,3 +281,78 @@ func (r *ReconcileKeycloak) ManageSuccess(instance *v1alpha1.Keycloak, currentSt
...
+	// Create Client which is working with PartialObjectMetadataList
+	metadataClient, err := metadata.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	// List all resources in cluster
+	// TODO: Investigate how to optimize this, it probably doesn't make sense to go through all resources
+	_, listResources, err := k8sClient.Discovery().ServerGroupsAndResources()
+	for _, resources := range listResources {
```
slaskawi, line 306: So the main question is how to search for all objects belonging to a specific Operator?

*Before I dig in, I highly recommend reading the [Kubernetes GC Manual](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).*

As mentioned in the documentation:

> In 1.8, Kubernetes automatically sets the value of ownerReference for objects created or adopted by ReplicationController, ReplicaSet, StatefulSet, DaemonSet, Deployment, Job and CronJob.

This means, those objects will be checked automatically here. We don't need to look at them. What we should search for are Secrets, ConfigMaps, Services, Ingress etc. All objects that may potentially be used by our Operator.

I think it's fine to iterate over those objects, but instead of checking the labels, we can inspect both `metadata.ownerReferences[0].controller` (needs to be true), `metadata.ownerReferences[0].kind` needs to be one of `Keycloak`, `KeycloakBackup`, `KeycloakUser`, `KeycloakRealm` or `KeycloakClient` CR and `metadata.ownerReferences[0].name` needs to be a valid CR name. Here's an output of a hand-written command:

```
kubectl get Secrets -o custom-columns="NAME:metadata.name,OWNING-REFERENCE-KIND:metadata.ownerReferences[0].kind,OWNING-REFERENCE-NAME:metadata.ownerReferences[0].name,CONTROLLER:metadata.ownerReferences[0].controller"

NAME                                OWNING-REFERENCE-KIND   OWNING-REFERENCE-NAME           CONTROLLER
builder-dockercfg-9ghcg             Secret                  builder-token-5c2kl             true
builder-token-5c2kl                 <none>                  <none>                          <none>
builder-token-bdnf5                 <none>                  <none>                          <none>
credential-example-keycloak         Keycloak                example-keycloak                true
default-dockercfg-c8xph             Secret                  default-token-99pbb             true
default-token-99pbb                 <none>                  <none>                          <none>
default-token-rzvzz                 <none>                  <none>                          <none>
deployer-dockercfg-5j2pm            Secret                  deployer-token-lsgv9            true
deployer-token-kw7bk                <none>                  <none>                          <none>
deployer-token-lsgv9                <none>                  <none>                          <none>
keycloak-db-secret                  Keycloak                example-keycloak                true
keycloak-operator-dockercfg-5vbkd   Secret                  keycloak-operator-token-jf7j5   true
keycloak-operator-token-jf7j5       <none>                  <none>                          <none>
keycloak-operator-token-z2wlx       <none>                  <none>                          <none>
rhsso-operator-dockercfg-pxs95      Secret                  rhsso-operator-token-2zkql      true
rhsso-operator-token-2zkql          <none>                  <none>                          <none>
rhsso-operator-token-5d9qr          <none>                  <none>                          <none>
sso-x509-https-secret               Service                 keycloak                        <none>
```

From the list the above, `keycloak-db-secret` meets our needs. The same step needs to be performed for all other objects. 

And once we managed to collect all the objects we need to inspect, we need to compare them with Secondary Resources list. If something is not on that list - it needs to be deleted. 

Does it help?

//cc @pb82 @davidffrench

slaskawi:
Closing it for now because of lack of progress. Please reopen once ready.

PR #177, opened 2020-04-30
KEYCLOAK-14035 Fix migration bug

mhajas:
## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/controller/keycloak/keycloak_migrations.go
+++ b/pkg/controller/keycloak/keycloak_migrations.go
@@ -28,7 +28,7 @@ func (i *DefaultMigrator) Migrate(cr *v1alpha1.Keycloak, currentState *common.Cl
 			desiredImage = model.RHSSOImage
 		}
 		log.Info(fmt.Sprintf("Performing migration from '%s' to '%s'", currentState.KeycloakDeployment.Spec.Template.Spec.Containers[0].Image, desiredImage))
-		deployment := findDeployment(desiredState)
+		deployment := findDeployment(&desiredState)
 		if deployment != nil {
 			log.Info("Number of replicas decreased to 1")
 			deployment.Spec.Replicas = &[]int32{1}[0]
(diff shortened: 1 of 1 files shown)


pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -178,17 +178,22 @@ func (r *ReconcileKeycloak) Reconcile(request reconcile.Request) (reconcile.Resu
...
 	// Perform migration if needed
+	var desiredState common.DesiredClusterState
+
 	migrator := NewDefaultMigrator()
 	desiredState, err = migrator.Migrate(instance, currentState, desiredState)
 	if err != nil {
 		return r.ManageError(instance, err)
 	}
 
+	// Skip Reconcilation of Keycloak if there is any migration needed
```
slaskawi, line 190: Here's how this should work:
- The `migrator` should operate on a processed `desiredState`. The idea is to modify it (the `desiredState`) to meet migration needs. 
- Implementing it as you proposed, will tie our hands for the future. A migrated version might require additional K8s objects (such as ConfigMaps or Secrets). As the `migrator` executes before the `reconciler` (and moreover, the `reconciler` backs off during the migration), those objects will not be created. This may lead to many problems in the future and very messy code. I'd like to avoid it.
- The `migrator` should obtain a copy of `desiredState` from the main loop (pass by value), find the `StatefulSet` object, change its number of replicas to 1 and pass it back to the main loop. 
- Once the `ActionRunner` executes it, the number of replicas should drop to 1 (if your cluster had 10 replicas, 9 of them should get into `terminating` state) and the `StatefulSet` rolling upgrade should kill the remaining one during the upgrade. 
- The next loop should increase the number of replicas back to 10. 

So it seems to me there some bug in the `migrator`, so that it doesn't return a modified `StatefulSet`. Would you mind to check it up?

mhajas: Thanks, @slaskawi. I see now. This was just an idea. I will incorporate your suggestions.

mhajas:
@slaskawi I think this is ready for review now.

PR #178, opened 2020-05-04
KEYCLOAK-13616 Setup for the product

slaskawi:
https://issues.redhat.com/browse/KEYCLOAK-13616

This Pull Request introduces a setup required for building both Keycloak and RHSSO from the same repo. OSBS (which is used by the product) requires additional YAML files (added by this commit). I also introduced `productize.sh` script, which replaces `Dockerfile` with the on prepared specially for OSBS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,15 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1 AS build-env
 
 RUN microdnf install -y git make golang
 
-ADD ./ /src
+COPY . /src/
 RUN cd /src && make code/compile
 RUN cd /src && echo "Build SHA1: $(git rev-parse HEAD)"
 RUN cd /src && echo "$(git rev-parse HEAD)" > /src/BUILD_INFO
 
 # final stage
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1
+
+##LABELS
 
 RUN microdnf update && microdnf clean all && rm -rf /var/cache/yum/*
 
--- a/container.yaml
+++ b/container.yaml
@@ -1,0 +1,10 @@
+compose:
+  pulp_repos: true
+go:
+  modules:
+    - module: github.com/keycloak/keycloak-operator
+image_build_method: imagebuilder
+platforms:
+  only:
+    - x86_64
+    - s390x
--- a/productize.sh
+++ b/productize.sh
@@ -1,0 +1,20 @@
+#!/bin/bash -ex
+
+LABELS=$(cat <<-END
+LABEL \
+com.redhat.component="redhat-sso-7-sso74-operator-rhel8-container"  \
+description="Red Hat Single Sign-On 7.4 Operator container image, based on the Red Hat Universal Base Image 8 Minimal container image" \
+summary="Red Hat Single Sign-On 7.4 Operator container image, based on the Red Hat Universal Base Image 8 Minimal container image" \
+version="7.4" \
+io.k8s.description="Operator for Red Hat SSO" \
+io.k8s.display-name="Red Hat SSO 7.4 Operator" \
+io.openshift.tags="sso,sso74,keycloak,operator" \
+name="rh-sso-7\/sso74-operator-rhel8" \
+maintainer="Red Hat Single Sign-On Team"
+END
+)
+
+sed -i 's/registry.access.redhat.com\/ubi8\/ubi-minimal:8.1/ubi8-minimal:8-released/' Dockerfile
+sed -i -e 's/FROM registry.access.redhat.com\/ubi8\/ubi-minimal:[0-9.]*/FROM ubi8-minimal:8-released/' \
+    -e "s/##LABELS/$LABELS/g" \
+    Dockerfile
--- a/content_sets.yaml
+++ b/content_sets.yaml
@@ -1,0 +1,9 @@
+ppc64le:
+  - rhel-8-for-ppc64le-baseos-rpms
+  - rhel-8-for-ppc64le-appstream-rpms
+s390x:
+  - rhel-8-for-s390x-baseos-rpms
+  - rhel-8-for-s390x-appstream-rpms
+x86_64:
+  - rhel-8-for-x86_64-baseos-rpms
+  - rhel-8-for-x86_64-appstream-rpms


slaskawi:
@ASzc @hmlnarik @mhajas FYI

stianst:
What is an downstream sync PR?! Changes go downstream, not upstream.

slaskawi:
@stianst We are syncing the code with changes in the downstream repo. Therefore, it's a downstream sync ;)

stianst:
Huh? Changes are applied upstream, then synced downstream. Not the other way around.

slaskawi:
@stianst I changes this to `OSBS sync`. Does this work for you?

stianst:
I take it OSBS requires the two extra yaml files to build the Operator? If so I would say a title "Add metadata files needed by OSBS" would be much better.

Lack of information in the title/description and just pointing to a large JIRA makes it really hard and time consuming to review things.

slaskawi:
@stianst Thanks for the comment!

I updated the commit as well as the description for this Pull Request. I also introduced the `productize.sh` script, which works a bit differently than what we agreed. I found putting labels a bit problematic (I had to index the lines), so in this proposal, I introduces a separate `Dockerfile.rhel8` file. I think this way it will be a bit less error prone.

slaskawi:
As per @hmlnarik request, I modified rendering product `Dockerfile` to a script. The result looks like the following:

```
$ cat Dockerfile 
FROM ubi8-minimal:8-released AS build-env

RUN microdnf install -y git make golang

COPY . /src/
RUN cd /src && make code/compile
RUN cd /src && echo "Build SHA1: $(git rev-parse HEAD)"
RUN cd /src && echo "$(git rev-parse HEAD)" > /src/BUILD_INFO

# final stage
FROM ubi8-minimal:8-released

LABEL com.redhat.component="redhat-sso-7-sso74-operator-rhel8-container"  description="Red Hat Single Sign-On 7.4 Operator on OpenJDK OpenShift container image, based on the Red Hat Universal Base Image 8 Minimal container image" summary="Red Hat Single Sign-On 7.4 Operator on OpenJDK OpenShift container image, based on the Red Hat Universal Base Image 8 Minimal container image" version="7.4" io.k8s.description="Operator for Red Hat SSO" io.k8s.display-name="Red Hat SSO 7.4 Operator" io.openshift.tags="sso,sso74,keycloak,operator" name="rh-sso-7/sso74-operator-rhel8" maintainer="Red Hat Single Sign-On Team"

COPY --from=build-env /src/BUILD_INFO /src/BUILD_INFO
COPY --from=build-env /src/tmp/_output/bin/keycloak-operator /usr/local/bin

ENTRYPOINT ["/usr/local/bin/keycloak-operator"]
```

Dockerfile
```diff
@@ -2,15 +2,15 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1 AS build-env
...
-ADD ./ /src
+COPY . /src/
 RUN cd /src && make code/compile
 RUN cd /src && echo "Build SHA1: $(git rev-parse HEAD)"
 RUN cd /src && echo "$(git rev-parse HEAD)" > /src/BUILD_INFO
 
 # final stage
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1
 
-RUN microdnf update && microdnf clean all && rm -rf /var/cache/yum/*
```
hmlnarik, line 13: Why is this step removed?

productize.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -ex
+
+LABELS=$(cat <<-END
+LABEL \
+com.redhat.component="redhat-sso-7-sso74-operator-rhel8-container"  \
+description="Red Hat Single Sign-On 7.4 Operator on OpenJDK OpenShift container image, based on the Red Hat Universal Base Image 8 Minimal container image" \
```
hmlnarik, line 6: Please remove reference to JDK since this is not relevant.

productize.sh
```diff
@@ -0,0 +1,18 @@
...
+io.k8s.description="Operator for Red Hat SSO" \
+io.k8s.display-name="Red Hat SSO 7.4 Operator" \
+io.openshift.tags="sso,sso74,keycloak,operator" \
+name="rh-sso-7\/sso74-operator-rhel8" \
+maintainer="Red Hat Single Sign-On Team"
+END
+)
+
+sed -i 's/registry.access.redhat.com\/ubi8\/ubi-minimal:8.1/ubi8-minimal:8-released/' Dockerfile
+sed -i "s/##LABELS/$LABELS/g" Dockerfile
```
hmlnarik, line 18:
```suggestion
sed -i -e 's/FROM registry.access.redhat.com\/ubi8\/ubi-minimal:[0-9.]*/FROM ubi8-minimal:8-released/' \
    -e "s/##LABELS/$LABELS/g" \
    Dockerfile
```

container.yaml
```diff
@@ -0,0 +1,11 @@
+compose:
+  pulp_repos: true
+go:
+  modules:
+    - module: github.com/keycloak/keycloak-operator
+image_build_method: imagebuilder
+platforms:
+  only:
+    - x86_64
+    - ppc64le
```
hmlnarik, line 10: Comment out ppc

Dockerfile
slaskawi: I realized this doesn't buy us anything.

In case of a CVE in UBI8, we get an updated package (with a CVE fix) via UBI8 image respin. Then, we need to respin our image to pick the fix up. We also don't rely on any upgraded package, so performing `microdnf update` doesn't help us in any sense. 

However, this has at least two pretty big drawbacks. At first it creates a new layer of RPMs, which may (depending on the amount of upgraded packages) increase the image size (moreover, the image size becomes unpredictable). It is far more beneficial to get those upgrades via a respun UBI8 image. The `microdnf update` procedure also has some (I guess very minimal, but still) impact on image integrity. We can be sure that pure UBI8 image has been verified by RHEL QE, but I wouldn't be so sure if they do verify the image with `microdnf upgrade` in it. I guess they do but without `microdnf upgrade` we take this question off the table.

So to sum it up - I think it does more harm than good. Since this is a new product release (and a tech preview), I think it's worth to verify my hypothesis and get rid of it.

productize.sh
slaskawi: Fixed.

productize.sh
slaskawi: Thanks for suggestion!

container.yaml
slaskawi: Removed.

slaskawi:
@hmlnarik Comments addressed.

Dockerfile
slaskawi: As per @hmlnarik request, re-introduced this piece.

PR #180, opened 2020-05-04
KEYCLOAK-14016 Support for testing operator as an image

vmuzikar:
## JIRA ID
[KEYCLOAK-14016](https://issues.redhat.com/browse/KEYCLOAK-14016)

## Additional Information
* Adds `test/e2e-image` make target to execute the tests with operator deployed to the cluster (as opposed to `test/e2e` which executes the operator locally while testing). This uses the `operator-sdk` CLI tool to automatically prepare the environment in a random namespace. From my PoV it doesn't make much sense to run this target in Travis as it uses an operator image and doesn't build and deploy the operator image.
* Fixes `WaitForOperatorDeployment` to use correct CR name.
* Updates docs on how to use ingress with Docker's K8s.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

## Additional Notes
I was not able to get `test/e2e-image` working with Ingress on my local machine. The operator doesn't seem to "communicate" with Ingress when deployed to the container (`test/e2e` works fine). Maybe I'm just doing something wrong. In any case, it's passing with OCP 4, so I'm not sure if it's even important for this to work locally as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,12 @@ test/e2e: cluster/prepare
 	# operator-sdk test  local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count" --namespace ${NAMESPACE} --up-local --debug --verbose ./test/e2e
 	go test -tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -mod=vendor ./test/e2e/... -root=$(PWD) -kubeconfig=$(HOME)/.kube/config -globalMan deploy/empty-init.yaml -namespacedMan deploy/empty-init.yaml -test.v -singleNamespace -parallel=1 -localOperator -test.timeout 0
 
+.PHONY: test/e2e-image
+test/e2e-image:
+	@echo Running tests with operator as an image in the cluster:
+	# Doesn't need cluster/prepare as it's done by operator-sdk. Uses a randomly generated namespace (instead of keycloak namespace) to support parallel test runs.
+	operator-sdk test local ./test/e2e --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count" --debug --verbose
+
 .PHONY: test/coverage/prepare
 test/coverage/prepare:
 	@echo Preparing coverage file:
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: keycloak-operator
           # Replace this with the built image name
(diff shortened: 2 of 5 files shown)


slaskawi:
Just thinking out laud - maybe this should our default e2e testing model? I believe it's absolutely fine to run `minikube docker-env && docker build . -t keycloak-operator:test`. This way the image should available for Minikube out of the box. Once the image build is done, we can trigger running the tests. 

Testing using an image has an additional advantage - we can test RBAC settings (and we can't do that with local deployment).

vmuzikar:
@slaskawi This will be the default mode in the test pipeline. ;)

slaskawi:
@vmuzikar That's a good point. I would like to make all the tests upstream and downstream the same. It simplifies a lot of stuff in the community.

What do you think @mhajas @davidffrench @pb82 about this idea (I mean always running tests against pre-built image, not a locally deployed Operator)?

mhajas:
> I was not able to get `test/e2e-image` working with Ingress on my local machine. The operator doesn't seem to "communicate" with Ingress when deployed to the container (`test/e2e` works fine). Maybe I'm just doing something wrong. In any case, it's passing with OCP 4, so I'm not sure if it's even important for this to work locally as well.

@vmuzikar the failures are caused by missing ingress support in the quay 9.0.2 image, it should be ok once the image is updated. There is some master tag in quay, not sure how and when it is built but it seems it works with that tag.

@slaskawi That's a great idea! I like it. This way the Travis build will catch more issues on the PR level. @vmuzikar Do you think you can try to play with this a little? Possibly to add it to this PR.

vmuzikar:
> @vmuzikar Do you think you can try to play with this a little? Possibly to add it to this PR.

@mhajas Sure, I'll try but I don't currently have bandwidth to do that. And I believe it's actually a bit out of scope of this PR, so created [KEYCLOAK-14063](https://issues.redhat.com/browse/KEYCLOAK-14063).

slaskawi:
@vmuzikar Thanks a lot!

slaskawi:
> the failures are caused by missing ingress support in the quay 9.0.2 image, it should be ok once the image is updated. There is some master tag in quay, not sure how and when it is built but it seems it works with that tag.

@mhajas @vmuzikar The `master` tag is rebuilt on every merge. It's a SNAPSHOT build. I believe we should be using it by default. @vmuzikar Perhaps you could check if the `master` tag works fine and if it does - please incorporate the change into this PR.

vmuzikar:
@slaskawi I can confirm what @mhajas said – it works for me with `master`. I've updated the PR as requested.

PR #181, opened 2020-05-05
Upstream/downstream Dockerfile proposal

slaskawi:
This Pull Request represents a proposal for hosting both upstream and downstream `Dockerfile`. As OSBS doesn't allow to use any other `Dockerfile` location, this is why the default needs to be set to the RHSSO one. Quay is much more liberal here and it allows to select `Dockerfile` location.


slaskawi:
This makes a lot of sense. Thanks for the input @stianst! Closing.

PR #186, opened 2020-05-08
KEYCLOAK-12307 TLS Termination configuration

slaskawi:
## JIRA ID

[KEYCLOAK-14095](https://issues.redhat.com/browse/KEYCLOAK-14095)
[KEYCLOAK-12307](https://issues.redhat.com/browse/KEYCLOAK-12307)

## Additional Information

This Pull Request changes the TLS configuration from Passthrough to Re-encrypt. In addition to that, it introduces a new `tlsTermination` parameter (in the `Keycloak` CR) that can be set to `passthrough` and fall back to previous behavior.

This change improves user experience and sacrifices a bit of security. As we agreed with @stianst, the safest way of running Keycloak is a Passthrough Route that sends all encrypted traffic to Keycloak. However, this option is quite hard to configure as the Signing Operator would need to create a valid, publicly accepted certificate to Keycloak. Much easier option is to configure this on an Ingress or Route. Moreover, OpenShift Installer has an option to install OpenShift with Let's Encrypt support out of the box. However, it's worth to mention, that OpenShift Router (HAProxy) will be performing re-encryption and if some malicious code gets there, all secret codes might be dumped somewhere. This is a security concern every administrator needs to consider.

Finally, this Pull Request doesn't touch Ingress. Ingress Controller uses, so called backside re-encryption (basically, it is configured to use HTTPS Backend) (see [link1](https://github.com/kubernetes/ingress-nginx/pull/519) [link2](https://github.com/kubernetes/kubernetes/issues/32928)). The Frontend TLS configuration is up to the user. The Operator doesn't reconcile it. 

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -44,6 +44,16 @@ spec:
                   description: If set to true, the Operator will create an Ingress
                     or a Route pointing to Keycloak.
                   type: boolean
+                tlsTermination:
+                  description: TLS Termination type for the external access. Setting
+                    this field to "reencrypt" will terminate TLS on the Ingress/Route
+                    level. Setting this field to "passthrough" will send encrypted
+                    traffic to the Pod. If unspecified, defaults to "reencrypt". Note,
+                    that this setting has no effect on Ingress as Ingress TLS settings
+                    are not reconciled by this operator. In other words, Ingress TLS
+                    configuration is the same in both cases and it is up to the user
+                    to configure TLS section of the Ingress.
+                  type: string
               type: object
             externalDatabase:
               description: "Controls external database settings. Using an external
--- a/pkg/model/keycloak_route.go
+++ b/pkg/model/keycloak_route.go
@@ -25,7 +25,7 @@ func KeycloakRoute(cr *kc.Keycloak) *v1.Route {
 				TargetPort: intstr.FromString(ApplicationName),
 			},
 			TLS: &v1.TLSConfig{
-				Termination: "passthrough",
+				Termination: getTLSTerminationType(cr),
 			},
 			To: v1.RouteTargetReference{
 				Kind: "Service",
@@ -42,7 +42,7 @@ func KeycloakRouteReconciled(cr *kc.Keycloak, currentState *v1.Route) *v1.Route 
 			TargetPort: intstr.FromString(ApplicationName),
 		},
 		TLS: &v1.TLSConfig{
-			Termination: "passthrough",
+			Termination: getTLSTerminationType(cr),
 		},
 		To: v1.RouteTargetReference{
 			Kind: "Service",
@@ -51,6 +51,13 @@ func KeycloakRouteReconciled(cr *kc.Keycloak, currentState *v1.Route) *v1.Route 
 	}
 
 	return reconciled
+}
+
+func getTLSTerminationType(cr *kc.Keycloak) v1.TLSTerminationType {
+	if cr.Spec.ExternalAccess.TLSTermination == kc.PassthroughTLSTerminationType {
+		return "passthrough"
+	}
+	return "reencrypt"
 }
 
 func KeycloakRouteSelector(cr *kc.Keycloak) client.ObjectKey {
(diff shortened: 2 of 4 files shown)


slaskawi:
@andymunro @pb82 @davidffrench You might be interested in this PR.

davidffrench:
Thanks @slaskawi . @pb82 this means we should be able to remove our additional route for keycloak in the integreatly-operator 🥳 

cc @pbrookes

slaskawi:
@davidffrench Yes, it took @stianst and myself a loooong time to decide on this.

pb82:
@slaskawi @davidffrench looks like it only can be `passthrough` or `reencrypt`: https://github.com/keycloak/keycloak-operator/pull/186/files#diff-5ddfce748972d078ca691ae71e4d08ffR56

We currently create a custom edge route.

slaskawi:
@pb82 Why do you use an edge-terminated route? If that's the Let's Encrypt case, the re-encrypt should do the job and it's much safer (as the traffic between HAProxy and the Pod is encrypted).

pb82:
@slaskawi I was wrong, we actually use re-encrypt. So this should be fine.

deploy/crds/keycloak.org_keycloaks_crd.yaml
```diff
@@ -44,6 +44,15 @@ spec:
                   description: If set to true, the Operator will create an Ingress
                     or a Route pointing to Keycloak.
                   type: boolean
+                tlsTermination:
```
stianst, line 47: From what I know OpenShift uses "reencrypt", "passthrough" and "edge". We should not support "edge" at all.

What does the value "router" mean? I actually don't understand if you mean "edge" or "reencrypt" here. I would just use the standard OpenShift names for it.

What's the default for this? I think we said "reencrypt" and I see in the code below that's what's used if the value is not set?! shouldn't the default behaviour be mentioned here?

Why do we not set the termination on Kubernetes? Do we not support setting up the Ingress at all?

slaskawi:
> I was wrong, we actually use re-encrypt. So this should be fine.

@pb82 Ok, that's great! Thanks Peter!

deploy/crds/keycloak.org_keycloaks_crd.yaml
slaskawi: > From what I know OpenShift uses "reencrypt", "passthrough" and "edge". We should not support "edge" at all.

And we don't. We support only re-encrypt and passthrough.

> What does the value "router" mean? I actually don't understand if you mean "edge" or "reencrypt" here. I would just use the standard OpenShift names for it.

I didn't want to use re-encrypt, since it's an implementation specific term. But after sleeping on it, I changed my mind.

> What's the default for this? I think we said "reencrypt" and I see in the code below that's what's used if the value is not set?! shouldn't the default behaviour be mentioned here?

Re-encrypt. Ok, I'll mention it.

> Why do we not set the termination on Kubernetes? Do we not support setting up the Ingress at all?

Yes, we do setup an ingress. However, frontend TLS termination requires creating a Secret with key/cert pair and referencing it in the Ingress. We don't have any way of generating those. They need to be supplied by a user. See the [documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls).

slaskawi:
@stianst Comments addressed and re-pushed.

pkg/model/keycloak_route.go
```diff
@@ -53,6 +53,13 @@ func KeycloakRouteReconciled(cr *kc.Keycloak, currentState *v1.Route) *v1.Route
 	return reconciled
 }
 
+func getTLSTerminationType(cr *kc.Keycloak) v1.TLSTerminationType {
+	if cr.Spec.ExternalAccess.TLSTermination == kc.ReencryptTLSTerminationType {
```
stianst, line 57: I'm not much of a go coder, but this doesn't look right ;)

slaskawi: No, that's fine :) The `ReencryptTLSTerminationType` is a type of a String.

stianst: Yes, but it says if the configured option is equal to reencrypt, then return passthrough?

slaskawi: ooh yes! You're absolutely right... last minute refactoring... arghhhhhh

slaskawi: Fixed.

mhajas: @slaskawi I think this should be actually the other way around. This way it defaults to `passthrough` and we need `reencrypt`. 

```suggestion
	if cr.Spec.ExternalAccess.TLSTermination == kc.PassthroughTLSTerminationType {
	    return "passtrhough"
	}
	return "reencrypt"
```

slaskawi:
@mhajas Should be fine now.

PR #187, opened 2020-05-08
[KEYCLOAK-14083] Run golangci-lint as part of pull request workflow

abstractj:
## [KEYCLOAK-14083](https://issues.redhat.com/browse/KEYCLOAK-14083)
This change introduce the usage of GolangCI linter with GH actions

## Additional Information
* [Example of GH actions working](https://github.com/arkham-labs/louketo-proxy/actions/runs/98490210)
* Output from the last run against Keycloak Operator repository - https://github.com/abstractj/keycloak-operator/actions/runs/99048680

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,27 @@
+name: Go
+
+on: [push, pull_request]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: go mod verify
+
+    - name: Build
+      run: |
+        go test ./...
+        go build -v .
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,15 @@
+name: Lint
+on: [push, pull_request]
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v0.1.7
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.26
+          github-token: "${{ secrets.GITHUB_TOKEN }}"


slaskawi:
@abstractj Just to make sure I fully understand how this works - with this change we will be inspecting every PR (and GH Actions will be adding comments if something is not right according to some of the linters) and on every SNAPSHOT build?

.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
+name: Go
```
stianst, line 1: Why is this workflow here? I though there would be a single lint workflow for this PR.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,15 @@
+name: Lint
+on: [push, pull_request]
```
stianst, line 2: Shouldn't this be only on pull_request?

abstractj: @stianst that's up to you and @slaskawi. The advantage of having this on push, is the fact that you can double check it if you push new tags. But honestly, that's your choice.

.github/workflows/go.yml
abstractj: @stianst having separate flows allows you to distinguish between the failures generated by the build and failures generate by the lack of formatting for example. Whether or not they should be in a single flow, again, it's your choice.
Most of the projects out there prefer to use it in separate flows.

abstractj:
> @abstractj Just to make sure I fully understand how this works - with this change we will be inspecting every PR (and GH Actions will be adding comments if something is not right according to some of the linters) and on every SNAPSHOT build?

Not really, because you need a bot to do that. Something which we had with GolangCI, probably Reviewdog can do this, but as @stianst mentioned, that's an extra service to worry about. What it does is to fail the build if things are broken. Here's an [example](https://github.com/arkham-labs/louketo-proxy/pull/2). You can also see how it looks like at the image below:

abstractj:
@slaskawi the build is failing if you look at https://github.com/abstractj/keycloak-operator/runs/657018502?check_suite_focus=true with the following message:

```
ok  	github.com/keycloak/keycloak-operator/pkg/model	0.049s
time="2020-05-08T17:02:46Z" level=fatal msg="Failed to create framework: failed to build the kubeconfig: invalid configuration: no configuration has been provided"
FAIL	github.com/keycloak/keycloak-operator/test/e2e	0.037s
?   	github.com/keycloak/keycloak-operator/version	[no test files]
```

Correct if I'm wrong, but I believe this can be fixed with https://github.com/keycloak/keycloak-operator/blob/master/Makefile#L147-L160. Does it make any sense?

abstractj:
Regards merge or not workflows. Once you guys reach a consensus, I can merge it into a single file or not.

slaskawi:
@abstractj Hey Bruno!

Thanks for the Pull Request. Unfortunately I misunderstood you and has an impression that the new linting machinery you proposed will issue comments on Pull Request, which is what I would love to have in this repo.

I'm not sure if it makes sense to hire Github Actions to perform this check. We already have `make code/lint` target, which performs the same thing. To me, in this very specific case it makes more sense to implement a separate stage in Travis to perform it (since you might be interested in test and linter results independently). 

Relying on Git Actions also has a pretty big drawback - you can not easily run it locally (opposed to `make code/lint` target). If you want to do it, you need to look at the Github Actions scripts and simulate running all the commands manually.

So to sum it up - I would prefer to have something that scans Pull Requests and adds comments to Pull Request. If this is not an option, I would prefer to rely on `make code/lint` rather than Github Actions.

abstractj:
@slaskawi @stianst absolutely np, feel free to close it.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v0.1.7
```
slaskawi, line 11: According to [the manual](https://github.com/golangci/golangci-lint-action), this should be `v1`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,15 @@
...
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.26
```
slaskawi, line 14: v.1.27

.github/workflows/lint.yml
abstractj: @slaskawi we already been using it in Louketo with no issues. But if you would like to use `v1`, that's no problem from me.

.github/workflows/lint.yml
abstractj: @slaskawi it's been a while since I merged this on Louketo, but I recall I had some issues with v1.27. But if that's what you would like, I don't mind.

Honestely I'd suggest to merge it, so you guys can iterate over it.

slaskawi:
Ok, let's give it a try

PR #197, opened 2020-05-20
KEYCLOAK-14261 Introduce RHSSO Init Container

slaskawi:
## JIRA ID

[KEYCLOAK-14261](https://issues.redhat.com/browse/KEYCLOAK-14261)

## Additional Information

This Pull Request introduces a new RHSSO Init Container.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/model/image_manager.go
+++ b/pkg/model/image_manager.go
@@ -11,13 +11,15 @@ const (
 	RHSSOImageOpenJDK     = "RELATED_IMAGE_RHSSO_OPENJDK"
 	RHSSOImage            = "RELATED_IMAGE_RHSSO"
 	KeycloakInitContainer = "RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER"
+	RHSSOInitContainer    = "RELATED_IMAGE_RHSSO_INIT_CONTAINER"
 	RHMIBackupContainer   = "RELATED_IMAGE_RHMI_BACKUP_CONTAINER"
 	PostgresqlImage       = "RELATED_IMAGE_POSTGRESQL"
 
 	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:9.0.2"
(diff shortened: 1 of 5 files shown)


pkg/model/image_manager.go
```diff
@@ -11,13 +11,15 @@ const (
...
 	RHSSOImage            = "RELATED_IMAGE_RHSSO"
 	KeycloakInitContainer = "RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER"
+	RHSSOInitContainer    = "RELATED_IMAGE_RHSSO_INIT_CONTAINER"
 	RHMIBackupContainer   = "RELATED_IMAGE_RHMI_BACKUP_CONTAINER"
 	PostgresqlImage       = "RELATED_IMAGE_POSTGRESQL"
 
 	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:9.0.2"
-	DefaultRHSSOImageOpenJ9      = "registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4-1"
-	DefaultRHSSOImageOpenJDK     = "registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4-1"
+	DefaultRHSSOImageOpenJ9      = "registry.redhat.io/rh-sso-7/sso74-openj9-openshift-rhel8:7.4"
```
mhajas, line 19: Do we want this tag? As the default, we should probably have some specific build instead of the latest. WDYT? The same for OpenJDK image.

slaskawi: I would like to preserve pinning (at least to some extends). With the `latest`, failures may come from nowhere....

mhajas: Yeah, that's exactly what I am afraid about with tag `7.4`. If I remember correctly tag `7.4` works as the latest 7.4 build. @hmlnarik could you confirm, please?

hmlnarik: That's my understanding, and per my understanding this should be here as suggested.

PR #203, opened 2020-05-31
Fix generation of keycloak client 

christianwoehrle:
... when  standardFlowEnabled and publicClient are set to false

## JIRA ID
KEYCLOAK-14088
KEYCLOAK-14134

## Additional Information
I have to create clients that have the fields standardFlowEnabled and publicCLient both set to false.
When the json for the respective keycloak api-call is created, these fields are missing due to the omitempty tag.
But keycloak used the default true for these values so the keycloak client get the wrong defaults.

Current implementation
```
	// True if Standard flow is enabled.
	// +optional
	StandardFlowEnabled bool `json:"standardFlowEnabled,omitempty"`
	// True if this is a public Client.
	// +optional
	PublicClient bool `json:"publicClient,omitempty"`
```

## Verification Steps

Apply this config:
```
kind: KeycloakClient
metadata:
  name: client-test
  labels:
    app: sso
spec:
  realmSelector:
    matchLabels:
      app: sso
  client:
    clientId: client-test
    secret: client-secret1
    clientAuthenticatorType: client-secret
    protocol: openid-connect
    publicClient: false
    directAccessGrantsEnabled: False
    standardFlowEnabled: false
    implicitFlowEnabled: False
    serviceAccountsEnabled: true
```
Without the fix the gui shows

Screenshot from 2020-05-31 15-39-17

With the fix:

Screenshot from 2020-05-31 15-38-18

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -14,7 +14,7 @@ spec:
     status: {}
   validation:
     openAPIV3Schema:
-      description: Keycloak is the Schema for the keycloaks API
+      description: Keycloak is the Schema for the keycloaks API .
       properties:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
@@ -29,7 +29,7 @@ spec:
         metadata:
           type: object
         spec:
-          description: KeycloakSpec defines the desired state of Keycloak.
+          description: KeycloakSpec defines the desired state of Keycloak. .
           properties:
             extensions:
               description: A list of extensions, where each one is a URL to a JAR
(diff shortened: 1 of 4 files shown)


slaskawi:
Let's give @mhajas a chance to look into this PR as well.

slaskawi:
@christianwoehrle Would you mind to rebase it? Once you're done, I'll integrate it.

christianwoehrle:
The linter errors are from files in test/e2e/*

I haven't changed these files with this PR. Do you see any actions necessary on my side?

slaskawi:
Integrated, thanks @christianwoehrle

PR #208, opened 2020-06-05
INTLY-8271 - skip backups if no instances to backup are found

pb82:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-14409

## Additional Information

When no matching realms or keycloaks were found for resources (clients, users, backups) then the reconciliation loop was skipped and an error was logged. This also prevented the finalizers on a resources from getting updated and thus, orphaned resources could no longer be deleted.

This change removes the error when no matches are found and instead treats that case as an error for backups exclusively (because backups actually require an instance to be present).

## Verification Steps

1. Deploy keycokak, a realm, a client, a user and a backup (you can use the examples)
2. Delete the realm CR
3. Delete the keycloak CR
4. Delete the user CR: this should still work and you should see a message in the operator logs saying the finalizer got removed.
5. Delete the client CR: this should still work and you should see a message in the operator logs saying the finalizer got removed
6. Delete the backup CR: this should work. No finalizers involved.

## Checklist:
- [ ] Verified by team member


go.mod
```diff
@@ -2,46 +2,25 @@ module github.com/keycloak/keycloak-operator
 
 require (
 	cloud.google.com/go v0.57.0 // indirect
-	github.com/Djarvur/go-err113 v0.1.0 // indirect
```
slaskawi, line 5: Would you mind to remove changes to go.mod and go.sum files?

hmlnarik:
@slaskawi Please introduce a KEYCLOAK JIRA. This JIRA needs to be referred from the commit (ref https://github.com/keycloak/keycloak-operator/blob/master/CONTRIBUTING.md#create-an-issue-in-keycloak-jira)

slaskawi:
@pb82 @hmlnarik Perhaps we can move the JIRA to Keycloak project? It will be the simplest thing to do.

go.mod
pb82: Done now, go.mod and go.sum reverted

pb82:
@slaskawi Yep, let me try to convert the JIRA

pb82:
@hmlnarik @slaskawi JIRA moved to Keycloak now: https://issues.redhat.com/browse/KEYCLOAK-14409

pkg/controller/keycloak/keycloak_controller.go
```diff
@@ -38,9 +38,9 @@ import (
 var log = logf.Log.WithName("controller_keycloak")
 
 const (
-	RequeueDelaySeconds      = 30 * time.Second
-	RequeueDelayErrorSeconds = 5 * time.Second
-	ControllerName           = "keycloak-controller"
+	RequeueDelay      = 30 * time.Second
```
lhanusov, line 41: @slaskawi In this file the timeout parameters for controller were refactored. The question is why the backup file was not handled same way? https://github.com/keycloak/keycloak-operator/pull/208/files#diff-a7a5480b67785a9ee86a8ef0ecf17e19

test/e2e/utils.go
```diff
@@ -135,6 +135,7 @@ func WaitForUserToBeReady(t *testing.T, framework *framework.Framework, namespac
 
 func WaitForSuccessResponseToContain(t *testing.T, framework *framework.Framework, url string, expectedString string) error {
 	return WaitForCondition(t, framework.KubeClient, func(t *testing.T, c kubernetes.Interface) error {
+		//golint:ignore
```
lhanusov, line 138: Is it OK to suppress the warning? It supposed to be a workaround?

pkg/controller/keycloak/keycloak_controller.go
slaskawi: Could you please rephrase your question? I think I don't understand it... Maybe I'm a bit "slow" today...

test/e2e/utils.go
slaskawi: Yes. I believe I removed this in my linter fix PR

pkg/controller/keycloak/keycloak_controller.go
lhanusov: In this file it was renamed RequeueDelaySeconds to RequeueDelay and RequeueDelayErrorSeconds to RequeueDelayError.
But in this file it remained as it is: https://github.com/keycloak/keycloak-operator/pull/208/files#diff-a7a5480b67785a9ee86a8ef0ecf17e19

slaskawi: ooh yes, you're probably right. @pb82 would you mind to issue a follow up PR?

pb82: The linter complained about the constant name containing `seconds` when the value was already adjusted to seconds. This was not the case for the constant in the backup controller. Not sure how useful this particular linter rule is though.

slaskawi: From what you're saying, code-wise we should be fine. However, it would be good to unify those variables.

slaskawi: I've just created a PR: https://github.com/keycloak/keycloak-operator/pull/210

PR #211, opened 2020-06-18
KEYCLOAK-13394 Add proxy settings

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-13394

## Additional Information

This Pull Request adds proper proxy settings and moves one test to the correct place.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/pkg/controller/keycloak/keycloak_reconciler_test.go
+++ b/pkg/controller/keycloak/keycloak_reconciler_test.go
@@ -633,8 +633,25 @@ func TestKeycloakReconciler_Test_No_Resources_Specified(t *testing.T) {
 	assert.Equal(t, len(postgresContainer.Resources.Limits), 0, "Limits-List should be empty")
 }
 
-func TestIsIP(t *testing.T) {
-	assert.True(t, model.IsIP([]byte("54.154.171.84")))
-	assert.False(t, model.IsIP([]byte("this.is.a.hostname")))
-	assert.False(t, model.IsIP([]byte("http://www.database.url")))
+func TestKeycloakReconciler_Test_Proxy_Settings(t *testing.T) {
+	// given
+	cr := &v1alpha1.Keycloak{}
+
+	currentState := common.NewClusterState()
+	reconciler := NewKeycloakReconciler()
+
+	// when
+	desiredState := reconciler.Reconcile(currentState, cr)
+
+	// then
+	// Expectation:
+	//    11) Keycloak StatefulSets
+	envs := desiredState[8].(common.GenericCreateAction).Ref.(*v13.StatefulSet).Spec.Template.Spec.Containers[0].Env
+	proxySet := false
+	for _, val := range envs {
+		if val.Name == "PROXY_ADDRESS_FORWARDING" && val.Value == "true" {
+			proxySet = true
+		}
+	}
+	assert.True(t, proxySet)
 }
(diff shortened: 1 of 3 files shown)


apapia:
This issue is blocking me so I'm hoping for a quick merge/release!

kfox1111:
@apapia unless your willing to build from source, I would look for an alternative solution to waiting for a release. They do not come out very fast. I ended up writing a mutating webhook to work around things.

pkg/model/rhsso_deployment.go
```diff
@@ -90,6 +90,10 @@ func getRHSSOEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 			Name:  "X509_CA_BUNDLE",
 			Value: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
 		},
+		{
+			Name:  "PROXY_ADDRESS_FORWARDING",
```
mhajas, line 94: From the Jira, it looks like this is not necessary for RHSSO. I haven't found any usage of this variable in our image, however, it is maybe necessary for the underlying container. @iankko @hmlnarik Do you know, whether it is necessary to add this ENV for RHSSO?

iankko: > From the Jira, it looks like this is not necessary for RHSSO. I haven't found any usage of this variable in our image, however, it is maybe necessary for the underlying container. @iankko @hmlnarik Do you know, whether it is necessary to add this ENV for RHSSO?

@mhajas It's (the `X509_CA_BUNDLE` variable) is needed just for the `*-x509-*` templates, or for the image itself only in the case, you want the RH-SSO https & Jgroups keystores & the RH-SSO truststore to be autogenerated. The code in question is [here](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/modules/sso/config/launch/setup/74/added/launch/openshift-x509.sh#L81) (there's corresponding counterpart for the RH-SSO 7.3 for OpenShift image too).

If the `X509_CA_BUNDLE` variable isn't defined, the image expects the user provide the definition of the RH-SSO https & JGroups keystores and the truststore. For non `*-x509-*` templates, this is done by setting corresponding necessary variables within them. [Here's the image message](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/templates/sso74-https.json#L22), detailing which of them need to be set the keystores & the trustore to work properly.

If it's not set at all (neither the `X509_CA_BUNDLE` variable, nor the `*_SECRET` variables are set), the SSL / TLS won't work for the image (the pod launch will fail with message about the impossibility to set up the keystores & truststore). Unsure if for the operator, the pre-creation of the keystores & truststores is handle somehow (@slaskawi do you know?) If there's such substitute functionality, we could probably use that one. If there isn't, we would need to utilize the `X509_CA_BUNDLE` (possibly allowing the user to specify custom CAs if necessary) variable as it's done in the images currently.

HTH, Jan

mhajas: @iankko Thanks for the explanation. Sorry I didn't mention it in my comment, but I was asking about the `PROXY_ADDRESS_FORWARDING` variable instead of `X509_CA_BUNDLE`.

apapia:
> @apapia unless your willing to build from source, I would look for an alternative solution to waiting for a release. They do not come out very fast. I ended up writing a mutating webhook to work around things.

Thanks @kfox1111.  I will fork the repo and build it myself for the time being.

pkg/model/rhsso_deployment.go
slaskawi: That's a good point. It's set to true by default (I mean the setting in the Undertow subsystem).

slaskawi: Removed.

slaskawi:
@mhajas Ready for another round of review.

PR #212, opened 2020-06-19
KEYCLOAK-14518 CRD cleanup

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-14518

## Additional Information

This Pull Request cleans our CRDs up.

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -129,13 +127,9 @@ var (
 	PhaseInitialising StatusPhase = "initialising"
 )
 
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
mhajas, line 132: @slaskawi Why are we getting rid of this? I found here: https://www.openshift.com/blog/kubernetes-deep-dive-code-generation-customresources that we need this:

> If you have tried to use CustomResources with client-go based on Kubernetes 1.8—some people might have had the pleasure already because they accidentally vendored a k8s.op/apimachinery of the master branch—you have hit the compiler error that the CustomResource type does not implement runtime.Object because DeepCopyObject() runtime.Object is not defined on your type. The reason is that in 1.8 the runtime.Object interface was extended with this method signature and hence every runtime.Object has to implement DeepCopyObject

However, Travis is happy about this, so this is probably not causing compilation error as written here ^^^ do you know why?

abstractj:
@slaskawi once you follow up on @mhajas, I can do the last round of review prior to merge.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
slaskawi: Let me double check this.

slaskawi: So currently our Kube dependency is too old. So let me revert those deep copy annotations - just in case.

PR #213, opened 2020-06-19
KEYCLOAK-14519 Development template and automation improvements

slaskawi:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-14519

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests

### DIFF
--- a/.github/workflows/commit-message.yml
+++ b/.github/workflows/commit-message.yml
@@ -1,0 +1,13 @@
+name: Commit Message
+on: [pull_request]
+jobs:
+  golangci:
+    name: commit-message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Check if JIRA ID is in the commit message
+        run: git log -n 1 --pretty=format:'%s' --abbrev-commit | grep -E 'KEYCLOAK-[0-9]+' > /dev/null
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,8 +4,9 @@
 ## Additional Information
 <!-- What/Why/How or any other context you feel is necessary.) -->
 
+<!-- (Add this section if applicable)
 ## Verification Steps
-<!--
+
 Add the steps required to check this change. Following an example.
 
 1. Go to `XX >> YY >> SS`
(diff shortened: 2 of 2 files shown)


.github/workflows/commit-message.yml
```diff
@@ -0,0 +1,13 @@
...
+  golangci:
+    name: commit-message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Check if JIRA ID is in the commit message
+        run: git log -n 1 --pretty=format:'%s' --abbrev-commit | grep -E 'KEYCLOAK-[0-9]+' > /dev/null
```
pb82, line 13: wouldn't this prevent community PRs since they usually start without an existing JIRA?

.github/workflows/commit-message.yml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  golangci:
+    name: commit-message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Check if JIRA ID is in the commit message
```
abstractj, line 12: @slaskawi could you please explain this step? When the Jira ID is missing the build will fail?

.github/workflows/commit-message.yml
slaskawi: Yes, I introduced a simple check that generates a failure (only in this one job) if the JIRA is missing. We've been chasing people to create JIRAs for their changes for ages. I'd like to make a step towards automating this.

As you probably spotted, out Github Pull Request template contains a paragraph where you should put your JIRA. In many cases, people tend to ignore it. I'd like to check if this will improve if they see a CI failure with a clear message. 

On the other hand, if you're modifying some README or fixing a tiny little typo somewhere, I don't really mind to merge a PR with this check red. However, it's a strong impulse to the PR author to put somewhere in the JIRA field as well as in the commit message.

.github/workflows/commit-message.yml
slaskawi: Yes, I believe @pb82 asked about exactly the same here (but using different words): https://github.com/keycloak/keycloak-operator/pull/213#discussion_r445503338. Let's continue discussion there.

.github/workflows/commit-message.yml
pb82: Ok, understand. If the policy is to have a JIRA for all but the most basic changes then this makes sense.

abstractj: @slaskawi thanks for clarifying it. But failing the build just because people forgot the Jira ID seems a bit misleading, assuming it does not have any relationship with tests or code.

I'd suggest another approach. If people forget to add the Jira ID, we automatically set the label `missing/jira` to the PR. But do not block the build.

slaskawi: It depends how you look at it. Here's an example I just created with missing JIRA:

Failing JIRA Golang Action

You clearly can see, that tests pass, linter pass but the only problem is with missing  JIRA. I could probably use slightly more suggestive name but this should be fine for now.

I believe my approach is much better than manually adding labels. First - you don't need to do it manually. Second - A Pull Request author receives an email with a failure build result. I really want to stay away from any manual work - it just don't scale.

slaskawi:
Thanks @abstractj ! Let me merge it and let's see how it goes. I think it's better to start collecting some feedback as soon as possible.

PR #215, opened 2020-06-22
KEYCLOAK-14559 updating ca bundle env variable with wildcard ca value

akoserwal:
## Jira

https://issues.redhat.com/browse/KEYCLOAK-14559

## Additional Information

Adding wildcard support for ca-cert configuration. This will allow configuration support for both Vanilla k8 & Openshift as the path varies var/run/secrets/kubernetes.io/serviceaccount/ca.crt (K8s) & /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt (Openshift).

Change is already done for `keycloak_deployment.go`. It was pending for `rhsso_deployment.go` this image release: https://github.com/jboss-container-images/redhat-sso-7-openshift-image/releases/tag/v7.4.0.GA-RHBZ-1813894-fix

https://github.com/jboss-container-images/redhat-sso-7-openshift-image/pull/125

### DIFF
--- a/pkg/model/rhsso_deployment.go
+++ b/pkg/model/rhsso_deployment.go
@@ -88,7 +88,7 @@ func getRHSSOEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 		},
 		{
 			Name:  "X509_CA_BUNDLE",
-			Value: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+			Value: "/var/run/secrets/kubernetes.io/serviceaccount/*.crt",
 		},
 	}
 


akoserwal:
@slaskawi: done

kfox1111:
Should this include /ca/*/*/* as well?

This would allow a secret named foo to be mounted at:
/ca/secret/foo/
and a configmap named foo:
/ca/configmap/foo/

?

slaskawi:
@kfox1111 I don't think so. In OpenShift, there are two CAs that should be taken into consideration here: `ca.crt` and `service-ca.crt`.

I noticed a couple of messages around CAs from you (some of them in [KEYCLOAK-13037](https://issues.redhat.com/browse/KEYCLOAK-13037)). Would you mind to describe your use case and post it to [Keycloak DEV](https://groups.google.com/forum/#!forum/keycloak-dev)? This way, we'll have a single place with most of the necessary information and it will be visible to the whole our team.

slaskawi:
Integrated, thanks @akoserwal

PR #216, opened 2020-06-23
KEYCLOAK-13112 Fixed client generation error and ImplicitFlow

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-13112

## Checklist:
- [ ] Verified by team member
- [ ] Automated Tests


slaskawi:
@mhajas @abstractj May I ask for a review?

slaskawi:
@abstractj Unfortunately, there's nothing we can do in the short time. We have a JIRA somewhere about creating OpenAPI Spec file from Keycloak REST interface. If I had such a spec, I could validate (or even generate) those CRDs from it. This would be an ultimate solution. However, @stianst mentioned, it won't be implemented anytime soon.

PR #218, opened 2020-06-24
KEYCLOAK-14573: Using new Red Hat registry for PG image

tux-o-matic:
## JIRA ID
KEYCLOAK-14573

## Additional Information
The Red Hat registry moved but was not used for the PostgreSQL image


tux-o-matic:
@slaskawi  [3 other images](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/image_manager.go#L19)  used by the operator are already pulled from registry.redhat.io

slaskawi:
@tux-o-matic Yes, that is correct. However, they are used only in RHSSO (product) profile. For installing RHSSO you need a Red Hat Subscription as well as Pull Secret for `registry.redhat.io`.

tux-o-matic:
Which is our case. In what cases is the Operator used without those 3 images?

slaskawi:
@tux-o-matic This is the Keycloak case (upstream). If you spin up a new Keycloak installation without specifying a profile - the Operator will spin up Keycloak (instead of RHSSO) for you. 

The bottom line - we use a single codebase for both RHSSO and Keycloak.

PR #220, opened 2020-06-25
KEYCLOAK-14327 Allow user to specify storage class

JiyeYu:
## JIRA ID
[KEYCLOAK-14327](https://issues.redhat.com/browse/KEYCLOAK-14327)

## Checklist:
- [x] Verified by team member
- [x] Comments where necessary
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloakbackups_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakbackups_crd.yaml
@@ -112,6 +112,10 @@ spec:
                 flag to true. Potentially, it will be possible to restore a single
                 backup multiple times."
               type: boolean
+            storageClassName:
+              description: Name of the StorageClass for Postgresql Backup Persistent
+                Volume Claim
+              type: string
           type: object
         status:
           description: KeycloakBackupStatus defines the observed state of KeycloakBackup.
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -150,6 +150,10 @@ spec:
               description: Profile used for controlling Operator behavior. Default
                 is empty.
               type: string
+            storageClassName:
+              description: Name of the StorageClass for Postgresql Persistent Volume
+                Claim
+              type: string
           type: object
         status:
           description: KeycloakStatus defines the observed state of Keycloak.
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -61,6 +61,9 @@ type KeycloakSpec struct {
 	// Specify Migration configuration
 	// +optional
 	Migration MigrateConfig `json:"migration,omitempty"`
+	// Name of the StorageClass for Postgresql Persistent Volume Claim
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
 }
 
 type DeploymentSpec struct {
--- a/pkg/apis/keycloak/v1alpha1/keycloakbackup_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloakbackup_types.go
@@ -25,6 +25,9 @@ type KeycloakBackupSpec struct {
 	// Selector for looking up Keycloak Custom Resources.
 	// +kubebuilder:validation:Required
 	InstanceSelector *metav1.LabelSelector `json:"instanceSelector,omitempty"`
+	// Name of the StorageClass for Postgresql Backup Persistent Volume Claim
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
 }
 
 // KeycloakAWSSpec defines the desired state of KeycloakBackupSpec.
--- a/pkg/apis/keycloak/v1alpha1/zz_generated.deepcopy.go
+++ b/pkg/apis/keycloak/v1alpha1/zz_generated.deepcopy.go
@@ -424,6 +424,11 @@ func (in *KeycloakBackupSpec) DeepCopyInto(out *KeycloakBackupSpec) {
 		*out = new(v1.LabelSelector)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.StorageClassName != nil {
+		in, out := &in.StorageClassName, &out.StorageClassName
+		*out = new(string)
+		**out = **in
+	}
 	return
 }
 
@@ -856,6 +861,11 @@ func (in *KeycloakSpec) DeepCopyInto(out *KeycloakSpec) {
 	in.KeycloakDeploymentSpec.DeepCopyInto(&out.KeycloakDeploymentSpec)
 	in.PostgresDeploymentSpec.DeepCopyInto(&out.PostgresDeploymentSpec)
 	out.Migration = in.Migration
+	if in.StorageClassName != nil {
+		in, out := &in.StorageClassName, &out.StorageClassName
+		*out = new(string)
+		**out = **in
+	}
 	return
 }
 
--- a/pkg/common/cluster_state.go
+++ b/pkg/common/cluster_state.go
@@ -524,6 +524,7 @@ func (i *ClusterState) readKeycloakBackupCurrentState(context context.Context, c
 	backupCr.Namespace = cr.Namespace
 	backupCr.Name = model.MigrateBackupName + "-" + BackupTime
 	backupCr.Spec.InstanceSelector = &labelSelect
+	backupCr.Spec.StorageClassName = cr.Spec.StorageClassName
 
 	KeycloakBackup := model.KeycloakMigrationOneTimeBackup(backupCr)
 	KeycloakBackupSelector := model.KeycloakMigrationOneTimeBackupSelector(backupCr)
--- a/pkg/controller/keycloak/keycloak_migrations.go
+++ b/pkg/controller/keycloak/keycloak_migrations.go
@@ -88,6 +88,7 @@ func oneTimeLocalBackup(cr *v1alpha1.Keycloak, currentState *common.ClusterState
 			MatchLabels: cr.Labels,
 		}
 		backupCr.Spec.InstanceSelector = &labelSelect
+		backupCr.Spec.StorageClassName = cr.Spec.StorageClassName
 
 		migrationBackupCR := common.GenericCreateAction{
 			Ref: model.KeycloakMigrationOneTimeBackup(backupCr),
--- a/pkg/controller/keycloak/keycloak_reconciler.go
+++ b/pkg/controller/keycloak/keycloak_reconciler.go
@@ -387,6 +387,7 @@ func (i *KeycloakReconciler) getKeycloakBackupDesiredState(clusterState *common.
 		MatchLabels: cr.Labels,
 	}
 	backupCr.Spec.InstanceSelector = &labelSelect
+	backupCr.Spec.StorageClassName = cr.Spec.StorageClassName
 
 	if clusterState.KeycloakBackup == nil {
 		// This happens before migration
--- a/pkg/model/postgresql_backup_persistent_volume_claim.go
+++ b/pkg/model/postgresql_backup_persistent_volume_claim.go
@@ -24,6 +24,7 @@ func PostgresqlBackupPersistentVolumeClaim(cr *v1alpha1.KeycloakBackup) *v1.Pers
 				Requests: v1.ResourceList{
 					v1.ResourceStorage: resource.MustParse(PostgresqlPersistentVolumeCapacity),
 				}},
+			StorageClassName: cr.Spec.StorageClassName,
 		},
 	}
 }
@@ -42,5 +43,8 @@ func PostgresqlBackupPersistentVolumeClaimReconciled(cr *v1alpha1.KeycloakBackup
 		Requests: v1.ResourceList{
 			v1.ResourceStorage: resource.MustParse(PostgresqlPersistentVolumeCapacity),
 		}}
+	if cr.Spec.StorageClassName != nil {
+		reconciled.Spec.StorageClassName = cr.Spec.StorageClassName
+	}
 	return reconciled
 }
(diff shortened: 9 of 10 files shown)


JiyeYu:
@slaskawi Do you know why this patch failed in CI test? 
The log gives nothing important information. Maybe we should run the CI again?

slaskawi:
@JiyeYu This failure usually indicates, that Keycloak deployment never gets up and running. I believe that's because Minikube creates a Default Storage Class. Does the test pass on your local machine?

JiyeYu:
Even if there is a default storage class, it should work. Locally I test this patch by creating Keycloak with or without StorageClassName in my minikube env.

However, I didn't do the e2e test locally because master branch also fails the e2e test in my local environment.

slaskawi:
@JiyeYu Running e2e tests locally is always the quickest option to debug a Travis failure. They pass without any problems on my local box as well as on @mhajas 's.

If you'd like to run them, please make sure you execute this two lines: https://github.com/keycloak/keycloak-operator/blob/master/Makefile#L163-L164

Everything has been documented here: https://github.com/keycloak/keycloak-operator/blob/master/docs/building.md#local-development

JiyeYu:
I am not sure the reason. Seems that in Minikube, Keycloak pod's Readiness probe and Liveness probe tend to fail easily.
When I use Minikube to test Keycloak locally, I will comment the Readiness probe and Liveness probe in file keycloak-deployment.go .

However, in my Kubernetes cluster environment which is created by kubespray, pod Keycloak-0's Readiness probe and Liveness probe all passed.

I am wondering whether this issue will cause the timeout in e2e test. Seems that CI is using Minikube to do the test.

Following is my case when I run the e2e test in my Minikube environment.

`$ minikube version
minikube version: v1.5.2
commit: 792dbf92a1de583fcee76f8791cff12e0c9440ad`

`$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-14T04:24:29Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}`

`$ kubectl describe pod keycloak-0 | grep fail
    Liveness:       exec [/bin/sh -c /probes/liveness_probe.sh] delay=30s timeout=22s period=30s #success=1 #failure=3
    Readiness:      exec [/bin/sh -c /probes/readiness_probe.sh] delay=40s timeout=22s period=30s #success=1 #failure=3
  Warning  Unhealthy  4m41s (x6 over 7m41s)  kubelet, minikube  Liveness probe failed:
  Normal   Killing    4m41s (x2 over 6m31s)  kubelet, minikube  Container keycloak failed liveness probe, will be restarted
  Warning  Unhealthy  3m24s (x7 over 7m18s)  kubelet, minikube  Readiness probe failed:`

slaskawi:
The readiness/liveness checks seem stable to me. But here's a pretty convenient way to debug them:

During the e2e, when Keycloak Pods are running but not ready, please check the following:
- `docker exec -it <pod id> bash`
- `$ bash -x ./probes/liveness_probe.sh`
- `$ bash -x ./probes/readiness_probe.sh`
- In the next terminal window, please check if the checks are really the problem: `kubectl describe pod keycloak-0`. Look at events section.

You may also try to modify `keycloak_probes.go` and try to find a problematic line. 

This would be very helpful for us. I've noticed some of the users report readiness check instability errors. However, I've never been able to reproduce them. 

And here's a bunch of a reference information about my environment:
```
➜  ~ minikube version                         
minikube version: v1.11.0
commit: 57e2f55f47effe9ce396cea42a1e0eb4f611ebbd
➜  ~ docker version                            
Client: Docker Engine - Community
 Version:           19.03.11
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        42e35e61f3
 Built:             Mon Jun  1 09:12:44 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.11
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       42e35e61f3
  Built:            Mon Jun  1 09:11:17 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

JiyeYu:
Thank you for your information. That's helpful!
I will build a new vm to run the e2e test later. 
Hope it can help me get rid of the interference.

JiyeYu:
@slaskawi 
I think I found the issue.
In the e2e test, the CR keycloak's creation will be interrupted by error:

> PersistentVolumeClaim "keycloak-postgresql-claim" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims

What happened is: 
PersistentVolumeClaim "keycloak-postgresql-claim has been successfully created by using the default storageclass "standard".
However, during the creation of CR keycloak, keycloak tries to change the PVC's storageclass again. (I am not sure the reason).
But the storageclass is immutable after creation. Then error happened.

This error is weird and only occurs in e2e test. 
I have tested manually by yaml file with/without parameter StorageClassName, and everything is fine. 

My solution for this issue is to change the e2e test case to pass the test.

JiyeYu:
> However, during the creation of CR keycloak, keycloak tries to change the PVC's storageclass again. (I am not sure the reason).

I think I find out the reason why keycloak tries to change the PVC's storageclass in e2e test. 

Because the PVC for e2e test is prepared before the creation of keycloak CR. [Here](https://github.com/keycloak/keycloak-operator/blob/master/test/e2e/minikube_issue_3129_workaround.go#L14)

slaskawi:
Let ma ask @mhajas to chime in.

Yes, testing this on Minikube will be very tricky as we have a Minikube Storage workaround in place. I think, the only way to do it is to create a `StorageClass` in Minikube that looks exactly the same as the default one.

mhajas:
@JiyeYu Do you have any idea how to avoid `StorageClassName: "standard"` in tests? It seems like it is necessary only for minikube so I would rather remove it from that configuration by default. Maybe we should do what @slaskawi suggested:

> I think, the only way to do it is to create a `StorageClass` in Minikube that looks exactly the same as the default one.

WDYT?

mhajas:
@JiyeYu I was wrong, the issue is present also in OCP cluster. It is failing with:

```
PersistentVolumeClaim "keycloak-postgresql-claim" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims
```

JiyeYu:
@mhajas Thank you for checking this.

> the issue is present also in OCP cluster. 

How did you test it in OCP cluster? Manually or through e2e test cases?  
I think this error only happens when PVC creating in advance, without using Keycloak's StorageClassName parameter.

JiyeYu:
@mhajas 
I tried OKD to test this patch.
I found there is no default StorageClass at the beginning. However, there are list of available pv to use.
However, if I create a StorageClass, and a PV using this StorageClass, 
then I can specify StorageClassName in Keycloak CR to use this new created PV.

So I think this patch also can be in OKD. 

```
$ oc version
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://127.0.0.1:8443
kubernetes v1.11.0+d4cacc0
```

```
$ oc get sc
NAME      PROVISIONER                    AGE
test      kubernetes.io/no-provisioner   10m
$ oc get pvc
NAME                        STATUS    VOLUME    CAPACITY   ACCESS MODES   STORAGECLASS   AGE
keycloak-postgresql-claim   Bound     pv123     1Gi        RWO            test           8m
$ oc get pv
...
pv0100    100Gi      RWO,ROX,RWX    Recycle          Available                                                                 46m
pv123     1Gi        RWO            Delete           Bound       keycloak/keycloak-postgresql-claim   test                     11m
```

JiyeYu:
> @JiyeYu Do you have any idea how to avoid `StorageClassName: "standard"` in tests? It seems like it is necessary only for minikube so I would rather remove it from that configuration by default. Maybe we should do what @slaskawi suggested:
> 
> > I think, the only way to do it is to create a `StorageClass` in Minikube that looks exactly the same as the default one.
> 
> WDYT?

To avoid this error, I think we should not create PVC for Keycloak in advance.
If we remove function "doWorkaroundIfNecessary", the PVC cannot be built up due to error "waiting for a volume to be created, either by external provisioner".
Seems that Minikube does not help to create a PV automatically.

So as a solution, I think creating a PV in advance is better than creating a PVC.
Then Keycloak CR will create a PVC by itself, and the PVC can be bound on the prepared PV.

mhajas:
> @mhajas Thank you for checking this.
> 
> > the issue is present also in OCP cluster.
> 
> How did you test it in OCP cluster? Manually or through e2e test cases?
> I think this error only happens when PVC creating in advance, without using Keycloak's StorageClassName parameter.

I tested with e2e, however, PVC should be created in advance only on minikube, on OCP the function `doWorkaroundIfNecessary` should not create PVC. 

> So as a solution, I think creating a PV in advance is better than creating a PVC.
Then Keycloak CR will create a PVC by itself, and the PVC can be bound on the prepared PV.

Could you update the tests to do it this way?

JiyeYu:
> > So as a solution, I think creating a PV in advance is better than creating a PVC.
> > Then Keycloak CR will create a PVC by itself, and the PVC can be bound on the prepared PV.
> 
> Could you update the tests to do it this way?

I'd like to do it. 
@slaskawi I think it's better to open another PR for implement of e2e test (creating PV in advance)
Otherwise the changes will be in chaos. I will create a new ticket as a continue.
Do you mind merging this pull request first?

slaskawi:
@JiyeYu Let's get @mhajas a chance to review it. Once he approves, I'll have a look. 

Just FYI - I might not be able to look into this today and on Thursday and Friday I'm on PTO. This PR will probably need to wait for Monday.

mhajas:
@JiyeYu To be honest, I don't really like the change to tests (addition of `StorageClassName`) by default. I think we are supposed to test it with default settings. If I understand it correctly, the `StorageClassName` should be used only if you want to use some custom Persistent volume, am I right? Therefore I would rather get rid of it within this PR. WDYT @JiyeYu @slaskawi?

JiyeYu:
> @JiyeYu To be honest, I don't really like the change to tests (addition of `StorageClassName`) by default. I think we are supposed to test it with default settings. If I understand it correctly, the `StorageClassName` should be used only if you want to use some custom Persistent volume, am I right? Therefore I would rather get rid of it within this PR. WDYT @JiyeYu @slaskawi?

I think the e2e test should have the same process as users do in real cases.
However, current e2e test is not perfect in my opinion. Because it changes resources' creation order.

The correct order should be 
Create Keycloak CR -> Create PVC (2nd resource) -> Create Postgresql Deployment (2nd resource)

However, the e2e test is doing 
Create PVC -> Create Keycloak CR -> Create Postgresql Deployment 

This order is the reason why this patch cannot pass the e2e test. 
In fact, this patch is quite an easy patch. It should not get involved into the e2e test's issue.

mhajas:
This behaviour is because of a bug in minikube. When we run tests against OCP we are running it in the order you mentioned. As I said if you are not using minikube, the function `doWorkaroundIfNecessary` should do nothing.

JiyeYu:
> As I said if you are not using minikube, the function `doWorkaroundIfNecessary` should do nothing.

@mhajas  Do you mind try it again? 
I also tested the e2e test in OKD. It went well.

I commented some useless code.
Here is the detailed code in file keycloak-operator/test/e2e/keycloaks_test.go

1. In function getKeycloakCR, we should remove line 32 and line 43
```
func getKeycloakCR(namespace string) *keycloakv1alpha1.Keycloak {
	//var StorageClass string = "standard"     **There is no standard sc in OKD, so we should comment it**
	return &keycloakv1alpha1.Keycloak{
		ObjectMeta: metav1.ObjectMeta{
			Name:      testKeycloakCRName,
			Namespace: namespace,
			Labels:    CreateLabel(namespace),
		},
		Spec: keycloakv1alpha1.KeycloakSpec{
			Instances:      1,
			ExternalAccess: keycloakv1alpha1.KeycloakExternalAccess{Enabled: true},
			Profile:        currentProfile(),
			//StorageClassName: &StorageClass,
		},
	}
}
```

2. In function prepareKeycloaksCR, we should remove line 55-58

```
func prepareKeycloaksCR(t *testing.T, f *framework.Framework, ctx *framework.TestCtx, namespace string) error {
	// err := doWorkaroundIfNecessary(f, ctx, namespace)  **Comment it out to confirm it will not work**
	// if err != nil {
	// 	return err
	// }

	keycloakCR := getKeycloakCR(namespace)
	err := Create(f, keycloakCR, ctx)
	if err != nil {
		return err
	}
```

mhajas:
@JiyeYu I did the same also before. I also set `watch` command on Keycloak CR, pods and PVCs and looked at what is created first. The CR was created first and then pods and PVC by the operator. What I did is just running two commands 
```
oc login .....
make test/e2e NAMESPACE=mhajas-operator-test
```

But maybe I am doing something wrong, we can wait for @slaskawi to try it as well on Monday.

mhajas:
I tried to debug it, and I can confirm that before this line: https://github.com/keycloak/keycloak-operator/blob/master/test/e2e/keycloaks_test.go#L59 no PVC was created.

JiyeYu:
> Do you have any idea how to avoid StorageClassName: "standard" in tests?
 
I do should avoid. I will first remove specify the StorageClassName: "standard".
Then the test in CI (Minikube env.) will fail. 
Maybe I should only wait Minikube to fix that confusing issue.

I will also do the e2e test again in OKD or other Kuberentes cluster to confirm if the same error happens or not.

JiyeYu:
@mhajas @slaskawi Good news!

This time I fixed my patch and solved this issue from the root.
I did not touch the e2e test cases and the CI test passed.
Please kindly review it again when you are free. Thank you.

slaskawi:
Let's give @mhajas a chance to review it first.

slaskawi:
Thanks @JiyeYu ! Integrated.

PR #222, opened 2020-07-01
KEYCLOAK-14627 Operator SDK v0.18.2

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-14627

## Additional Information

This Pull Request updates Operator SDK to v0.18.2. This change is required as a prerequisite to support the new Bundle Format.

Wow!... what a painful update....

## Checklist:
- [x] Automated Tests

This Pull Request contains 2 commits - the upgrade itself and vendoring changes. Please select proper commit for review.


.travis.yml
```diff
@@ -24,8 +24,7 @@ before_install:
   - sudo apt-get install -y conntrack
   - make setup/travis
 
-# Since there's no persistence sharing, we need to run everything in a single stage
-install: make $TESTS
+install: make ${TESTS} NAMESPACE=default
```
slaskawi, line 27: At this point, I'm almost certain our goveralls need to be rebuild. I'm not sure how, but there must be an option to collect multiple coverage reports in their service.

Makefile
```diff
@@ -49,28 +50,22 @@ test/unit:
...
-	# However, this way we will have a consistent way of running tests on Travis and locally. The downside
-	# is that Operator testing harness downloads things manually using `go mod` when executing the tests.
-	# Here is a corresponding Operator SDK call:
-	# operator-sdk test  local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count" --namespace ${NAMESPACE} --up-local --debug --verbose ./test/e2e
-	go test -tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -mod=vendor ./test/e2e/... -root=$(PWD) -kubeconfig=$(HOME)/.kube/config -globalMan deploy/empty-init.yaml -namespacedMan deploy/empty-init.yaml -test.v -singleNamespace -parallel=1 -localOperator -test.timeout 0
+test/e2e: setup/operator-sdk
+	@echo Running e2e local tests:
+	operator-sdk test local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0" --operator-namespace $(NAMESPACE) --up-local --debug --verbose ./test/e2e
```
slaskawi, line 55: Unfortunately, there is no way to run this from `go test`. There's no switch to tell the testing harness to use a specific namespace. Those options are only exposed in `operator-sdk test` command.

slaskawi:
@pb82 @davidffrench Perhaps you'd be interested to look into this as well? Upgrading to 18.x.y was very painful...

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
```diff
@@ -10,144 +10,144 @@ spec:
...
-                        type: string
-                      values:
-                        description: values is an array of string values. If the operator
-                          is In or NotIn, the values array must be non-empty. If the
-                          operator is Exists or DoesNotExist, the values array must
-                          be empty. This array is replaced during a strategic merge
-                          patch.
-                        items:
+  versions:
+  - name: v1alpha1
```
davidffrench, line 14: Do you think the version should be bumped with this switch to `v1`?

The same applies to the other CRDs as well.

hack/keycloak-postgresql-pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
davidffrench, line 1: Should this have been committed?

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
slaskawi: This is something that has been introduced in Operator SDK 18.0.0 (see the [release notes](https://github.com/operator-framework/operator-sdk/releases/tag/v0.18.0)). This should be fine for K8s `1.16+`. Since we use `alpha` for our CRDs, I think that's a safe change. WDYT?

hack/keycloak-postgresql-pvc.yaml
slaskawi: Yes, it should be added earlier. This might be helpful when applying a workaround for https://github.com/kubernetes/minikube/issues/3129

slaskawi:
I'll also update the Operator code with suggestions in https://sdk.operatorframework.io/docs/migration/version-upgrade-guide/#v016x. Thanks @davidffrench for adding this document. It's super useful!

hack/keycloak-postgresql-pvc.yaml
davidffrench: Thanks @slaskawi

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
davidffrench: @slaskawi More operators are going to be going through the upgrade of CRDs to using `v1`. I believe some fields in the spec may have been removed. 

https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1beta1/types.go vs https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go

Since the Keycloak CRDs are `v1alpha1`, you can make as many changes as you want. Might still be worth bumping the version depending on the diff between the above links.

slaskawi: Let me ask @stianst for opinion here.

slaskawi:
Updated according to the upgrade guide: https://sdk.operatorframework.io/docs/migration/version-upgrade-guide/#v016x

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
slaskawi: Let me put this on hold for a moment. It just came to me that `apiVersion` change might be problematic from OLM upgrade point of view. I asked Operator SDK Team to clarify this.

slaskawi:
Putting on hold till I clarify that OLM upgrades will work fine with this change.

deploy/crds/keycloak.org_keycloakbackups_crd.yaml
slaskawi: We had a chat internally about this migration. There are two important aspects:
- The API Server will automatically and transparently convert CRDs between `v1beta1` and `v1`. There should be no issues from this perspective.
- OpenShift 4.3 and 4.4 (that we use for the testing in the product) requires `v1beta1`. This means it's our lower bound and we simply can not migrate to `v1` at this stage.

@davidffrench @pb82 This might be also important to you. You will probably have a similar conversation about upgrades soon.

//cc @abstractj @vmuzikar @ASzc

davidffrench: Perfect, thank you @slaskawi , this is fantastic information. I was following your thread on the operator framework mailing list as well.

slaskawi:
I reverted our CRDs to `v1beta1` and updated the `make code/gen` script to keep the proper CRD version.

slaskawi:
@ASzc @abstractj @mhajas May I ask for a review?

PR #226, opened 2020-07-07
KEYCLOAK-14590 Update productize.sh for exploded sources

ASzc:
Needed for keycloak/keycloak-pipeline#291

### DIFF
--- a/productize.sh
+++ b/productize.sh
@@ -17,5 +17,7 @@ END
 sed -i \
     -e 's/registry.svc.ci.openshift.org\/openshift\/release:golang-1.13/openshift\/golang-builder:1.13/' \
     -e 's/FROM registry.access.redhat.com\/ubi8\/ubi-minimal:[0-9.]*/FROM ubi8-minimal:8-released/' \
+    -e 's/COPY . /COPY keycloak-operator-*.tar.gz /' \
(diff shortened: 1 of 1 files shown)


productize.sh
```diff
@@ -17,5 +17,7 @@ END
 sed -i \
     -e 's/registry.svc.ci.openshift.org\/openshift\/release:golang-1.13/openshift\/golang-builder:1.13/' \
     -e 's/FROM registry.access.redhat.com\/ubi8\/ubi-minimal:[0-9.]*/FROM ubi8-minimal:8-released/' \
+    -e 's/COPY . /COPY keycloak-operator-*.tar.gz /' \
+    -e 's,RUN cd /src ,RUN cd /src \&\& tar -x --strip-components=1 -f keycloak-operator-*.tar.gz ,' \
```
slaskawi, line 21: Wow, I've never seen `sed` syntax with `,`. Just double checking if this is what you need here (and is not a typo)?

ASzc: Technically you can use any character you want (certainly any symbol character anyway) after the s, and that becomes the delimiter character. In this case I'm dealing with paths in the pattern, so I need something other than `/` as the delimiter to avoid the pain of escaping those.

For example:

```bash
[aszczucz@tyrfing ~]$ sed -e 's,/home/,/etc/,' <<<"/home/alex/"
/etc/alex/
```

slaskawi: Cool! I didn't know that! Thanks for clarifying!

PR #229, opened 2020-07-13
KEYCLOAK-14316 Fix "assignment to entry in nil map" error

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-14316

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/model/database_secret.go
+++ b/pkg/model/database_secret.go
@@ -16,13 +16,13 @@ func DatabaseSecret(cr *v1alpha1.Keycloak) *v1.Secret {
 				"app": ApplicationName,
 			},
 		},
-		StringData: map[string]string{
-			DatabaseSecretUsernameProperty: PostgresqlUsername,
-			DatabaseSecretPasswordProperty: cr.ObjectMeta.Name + "-" + GenerateRandomString(PostgresqlPasswordLength),
+		Data: map[string][]byte{
+			DatabaseSecretUsernameProperty: []byte(PostgresqlUsername),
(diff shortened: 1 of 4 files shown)


pkg/model/database_secret.go
```diff
@@ -36,21 +36,25 @@ func DatabaseSecretSelector(cr *v1alpha1.Keycloak) client.ObjectKey {
 
 func DatabaseSecretReconciled(cr *v1alpha1.Keycloak, currentState *v1.Secret) *v1.Secret {
 	reconciled := currentState.DeepCopy()
+	if reconciled.Data == nil || len(reconciled.Data) == 0 {
+		reconciled.Data = make(map[string][]byte)
+	}
+
 	// K8s automatically converts StringData to Data when getting the resource
```
chlunde, line 43: This comment is now outdated

Also, I think it is now missing base64 encoding.

The stringData field is a write-only convenience field. It is never output when retrieving Secrets.

slaskawi: That's definitely an old comment! 

Currently we use the `StringData` approach. This has one big drawback - if you put anything there, it's being translated into `Data` and K8s wipes `StringData` out. So, there's a different place to read from and write to. This is pretty error prone and the JIRA attached to this PR is a good example.

I'll remove this obsolete comment. Thanks for spotting it!

slaskawi:
@abstractj May I ask you to review and merge this one (already approved by @ASzc).

PR #231, opened 2020-07-15
KEYCLOAK-14782 - Support of additional environment variables

yannickguth:
## JIRA ID
[KEYCLOAK-14782](https://issues.redhat.com/browse/KEYCLOAK-14782)

## Additional Information
In the current version of the Keycloak Operator some environment variables are hardcoded. For example, CACHE_OWNER_COUNT.
I would like to have the possibility to select or overwrite these hardcoded variables myself and also to add other variables that might be used by own SPIs.

Therefore I have extended the Keycloak Spec with the possibility to add environment variables or configmaps.

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -154,6 +154,9 @@ spec:
               description: Name of the StorageClass for Postgresql Persistent Volume
                 Claim
               type: string
+            env:
+              description: Environment variables to add or overwrite the hard-coded
+              type: object
           type: object
         status:
           description: KeycloakStatus defines the observed state of Keycloak.
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -154,8 +154,11 @@ spec:
               description: Name of the StorageClass for Postgresql Persistent Volume
                 Claim
               type: string
+            envFrom:
+              description: Config- or secretMaps for additional environment variables (does not overwrite the directly set environment variables)
+              type: object
             env:
-              description: Environment variables to add or overwrite the hard-coded
+              description: Environment variables to add or overwrite the hard-coded ones
               type: object
           type: object
         status:
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -65,6 +65,9 @@ type KeycloakSpec struct {
 	// +optional
 	StorageClassName *string `json:"storageClassName,omitempty"`
 
+	// Additional configmaps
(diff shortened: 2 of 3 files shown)


slaskawi:
In the past we used to reject Pull Requests like this (see [here](https://github.com/keycloak/keycloak-operator/pull/165#issuecomment-618256163) as an example).

However, you have a point saying that custom providers deployed by the Operator often need custom configuration. This needs to be passed somehow into the Keycloak Pod. From this angle, this has a lot in common with [KEYCLOAK-12677](https://issues.redhat.com/browse/KEYCLOAK-12677) (which also has an ongoing discussion at [Keycloak DEV Group](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/keycloak-dev/mex1RqhlvBI/PaFHNYJnBQAJ)). 

I believe, both mounting ConfigMaps and adding env. variables to Keycloak Pod could (or even should) be addressed as the feature. Looking from this angle, this Pull Request probably needs to be rejected as mounting configmaps and adding custom env. variables should happen in the same section of the CRD. We already scheduled looking into this in Q4 this year. 

@stianst @mhajas @yannickguth @abstractj WDYT?

slaskawi:
Let me put this on hold for now. The implementation needs to be tightly coupled with [KEYCLOAK-12677](https://issues.redhat.com/browse/KEYCLOAK-12677) and probably should be implemented in the same Pull Request. 

There are number of open questions here including:
- Can I override an env. variable already set by the Operator?
- Should we use validation for those env. variables? 
- Is there any better way to deliver variables to the running image? Perhaps a ConfigMap is a better concept?

slaskawi:
I've just created a design for this feature. Please see https://mail.google.com/mail/u/0/#inbox/FMfcgxwKhqrMCvvmnKMBJzblKWZmVNfh

We would be happy if you decided to help us out and implement some parts of the Custom Configuration for the Operator. This however need to be done using slightly different CR structure than you proposed. 

Let me close this Pull Request. You may reopen once you modify the structure.

yannickguth:
I'm glad that my original proposal was discussed so intensively and now actually gets the chance to be included in the project. I will try to start with your proposed implementation as soon as possible.

slaskawi:
@yannickguth Yes, it took us quite a while to set the ground for this feature. Thanks for creating the initial version of it!

As the matter of fact, I already started working on it: https://github.com/keycloak/keycloak-operator/pull/255

Today I plan to finish the testing up and it should be ready for review. It would be great to have your feedback on it.

Kampe:
Theres a grand total of zero examples of using environment variables.   Given this ticket, can this please add that to a README somewhere?

slaskawi:
@Kampe The lack of examples is somewhat "by design". Those are experimental settings and can be wiped out without a notice. They need to be treated as a last resort and used only when there's no other solution. We don't want our Community to rely on them - we want to properly support their use cases.

PR #232, opened 2020-07-17
explicit set the FailureThreshold for the readiness and liveness probes

wei-lee:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-14814

## Verification Steps

see JIRA description

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/model/keycloak_deployment.go
+++ b/pkg/model/keycloak_deployment.go
@@ -19,6 +19,7 @@ const (
 	//10s (curl) + 10s (curl) + 2s (just in case)
 	ProbeTimeoutSeconds         = 22
 	ProbeTimeBetweenRunsSeconds = 30
+	ProbeFailureThreshold       = 10
 )
 
 func GetServiceEnvVar(suffix string) string {
@@ -328,6 +329,7 @@ func livenessProbe() *v1.Probe {
 		InitialDelaySeconds: LivenessProbeInitialDelay,
 		TimeoutSeconds:      ProbeTimeoutSeconds,
 		PeriodSeconds:       ProbeTimeBetweenRunsSeconds,
+		FailureThreshold:    ProbeFailureThreshold,
 	}
 }
 
@@ -345,5 +347,6 @@ func readinessProbe() *v1.Probe {
 		InitialDelaySeconds: ReadinessProbeInitialDelay,
 		TimeoutSeconds:      ProbeTimeoutSeconds,
 		PeriodSeconds:       ProbeTimeBetweenRunsSeconds,
+		FailureThreshold:    ProbeFailureThreshold,
 	}
 }


wei-lee:
@slaskawi Can you review this please? Thanks.

slaskawi:
Thanks for this PR, @wei-lee !

Does this change improve anything in the production clusters (from RHMI for example)? 

From my experience, long bootstrap times may happen when there's not enough CPU assigned to the Keycloak Pods. So I wonder if addressing this by increasing the failure threshold is the right thing to do.

wei-lee:
@slaskawi I didn't see the problem in production clusters. I see it in the RHPDS clusters.

> From my experience, long bootstrap times may happen when there's not enough CPU assigned to the Keycloak Pods. So I wonder if addressing this by increasing the failure threshold is the right thing to do.

Yeah, it probably isn't the best option. But I don't know what other way we can fix it. Should we increase the resource request for keycloak then?

slaskawi:
@wei-lee I think it's worth checking if this improves the situation. Please keep me posted.

On a side note - this PR looks good to me, but I would prefer to make sure people are not using too low CPU settings and it will hurt the performance (a lot!). Issuing tokens is a CPU sensitive thing.

wei-lee:
@slaskawi I can certainly check on. But I was just using the default configurations for CPU resources. So we are not seeing this issue anywhere else? If so it could be something specific to RHPDS.

slaskawi:
@wei-lee That's very weird. RHPDS is my main test environment and I've never seen problems like this. So I would love to see what will happen if you set the resources. Could you please tell me what values do you use today?

wei-lee:
@slaskawi I tried with `cpu: 0.5 memory: 1Gi` but having the same issue. For some reason I can't increase the cpu value to `1` due to some resource limit, but I can't find where the limit is set.

This is the cluster I have using: https://console-openshift-console.apps.cluster-demo-9739.demo-9739.example.opentlc.com/. You can login as the `opentlc-mgr` user (if you don't have the password ping me). If you check the keycloak namespace you will see the problem there.

slaskawi:
I did a few experiments within your cluster (`slaskawi` namespace). For the testing purpose I used Requests = Limits (which gives you the best QoS):

- Requests = Limits => CPU = 50m - everything failed. It didn't even "try" to start. Java is not Go :D
- Requests = Limits => CPU = 500m - fails to start and keeps restarting. Still too little.
- Requests = Limits => CPU = 750m - fails to start and keeps restarting. Still too little.
- Requests = Limits => CPU = 1 - Started in 2m

Generating KeyStore and TrustStore takes the largest portion of time. This is somewhat expected (or at least shouldn't be a surprise). 

I suspect, many community users will use CPU below `1`. `500m` sounds pretty realistic and Keycloak should boot up with such constrains. Therefore, I'd like to get this PR in. 

@wei-lee May I ask you to remove changes in `go.sum`? They are unrelated to this PR.

wei-lee:
@slaskawi sure, no problem.

wei-lee:
@slaskawi I have addressed your comment. Thanks.

slaskawi:
Integrated, thanks @wei-lee

PR #239, opened 2020-07-28
KEYCLOAK-14931: Remove `cluster.local` from DNS names in the deployment genertor.

briantopping:
Kube-DNS is configured with `ndots: 5`, which means that local names will be iteratively retried with five root names of the FQDN removed.

This fixes clusters that have DNS names other than `cluster.local`.

## Verification Steps

1. Create a cluster with a specific cluster name such as `c1.example.com`
1. Install Keycloak-Operator, notice operator generates a `Deployment` that has `cluster.local` in it's DNS names. These DNS names are incorrect and the deployment fails.
1. Apply patch and try again. `kube-dns` is configured with `ndots: 5`, which solves the problem.

### DIFF
--- a/pkg/model/keycloak_deployment.go
+++ b/pkg/model/keycloak_deployment.go
@@ -67,7 +67,7 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 		},
 		{
 			Name:  "DB_ADDR",
-			Value: PostgresqlServiceName + "." + cr.Namespace + ".svc.cluster.local",
+			Value: PostgresqlServiceName + "." + cr.Namespace,
 		},
 		{
 			Name:  "DB_DATABASE",
@@ -106,7 +106,7 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 		},
 		{
 			Name:  "JGROUPS_DISCOVERY_PROPERTIES",
-			Value: "dns_query=" + KeycloakDiscoveryServiceName + "." + cr.Namespace + ".svc.cluster.local",
+			Value: "dns_query=" + KeycloakDiscoveryServiceName + "." + cr.Namespace,
 		},
 		// Cache settings
 		{


slaskawi:
This in very interesting stuff, @briantopping !

I went through [this excellent article](https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html) that describes the mechanics (and impact) of `ndots: 5` change. Here's the relevant part:

> ndots:n
>
> sets a threshold for the number of dots which must appear in a name before an initial absolute query will be made. The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it.
>
> This means that if ndots is set to 5 and the name contains less than 5 dots inside it, the syscall will try to resolve it sequentially going through all local search domains first and - in case none succeed - will resolve it as an absolute name only at last.

I've just checked what OpenShift (which is equivalent to K8s in this case) injects into the `/etc/resolv.conf` of the pods:

```
$ cat /etc/resolv.conf
search keycloak.svc.cluster.local svc.cluster.local cluster.local eu-central-1.compute.internal
nameserver 172.30.0.10
options ndots:5
```

So it seem it also appends `svc.cluster.local` as well. Having this in mind, I believe we can change the query name (in all parts of the code) to this: 

    dns_query=" + KeycloakDiscoveryServiceName + "." + cr.Namespace

As you probably noticed, I also removed the `svc` part, which should also be included in the search list. 

If you agree with me, would you mind to update the Pull Request (and check if this works correctly)?

//cc @rigazilla - you might want to include this change in the Infinispan Operator as well. Possibly `DNS_PING` stacks should also be updated.

briantopping:
Hi @slaskawi, I'm glad you're having fun with it! I also did when I first discovered it. 😀

Yes, I agree with your change. I was going from memory on the search path order and had it backwards. Your proposal is the shorter path. 

Glad it was helpful! Change is pushed.

slaskawi:
Integrated, thanks @briantopping

briantopping:
Great, thanks team!! Glad to have been able to give a small amount to such a great project.

PR #241, opened 2020-08-05
KEYCLOAK-15014 Add scopes to realm and client

chlunde:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-15014

## Additional Information
Supporting scopes is nice to be able to add additional claims to JWT tokens.

**This PR builds on other PRs, only the last commit is relevant for this change**

## Verification Steps

The REST API is tested using e2e import/export.

Manual testing is possible:

1. Create a CR with a custom scope adding a static attribute and all other required mappers
2. Log in and verify the claims

## Checklist:
- [x] Automated Tests


slaskawi (changes requested):
@chlunde Again, I really like the code changes but the same review comment applies here as well: https://github.com/keycloak/keycloak-operator/pull/230#pullrequestreview-471543004

chlunde:
@slaskawi done. I did a slight re-ordering of the original fields to avoid conflict after merging the other pending PR.

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
```diff
@@ -160,6 +160,54 @@ spec:
                     - alias
                     type: object
                   type: array
+                clientScopes:
+                  description: Client scopes
+                  items:
+                    description: https://www.keycloak.org/docs-api/10.0/rest-api/index.html#_clientscoperepresentation
```
slaskawi, line 166: We probably should avoid this kind of descriptions (although I think I missed this in the previous PR I reviewed). This either should be empty or have some meaningful description.

slaskawi (changes requested):
Only one small thing to be corrected. The other bits are fine.

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
chlunde: Sorry, that was intented as a comment in the source, not CRD docs.

I think the user docs should be exported from keycloak somehow.

For now we can adapt this text I guess: https://www.keycloak.org/docs/latest/server_admin/#core-concepts-and-terms

slaskawi: We have a documentation for our REST resources here: https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_realmrepresentation (you may find other resources, such as `ClientRepresentations` or `UserRepresentation` there as well).

slaskawi:
Hey @chlunde ! Do you plan to work on this PR in the near future? I'd like to get it merged.

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
chlunde: I'll just leave it empty as you suggest:
* I don't think the current overall CRD documentation is helpful enough overall currently for anyone to use it alone
* I think all "hand written" comments would be removed (and come from OpenAPI) when https://issues.redhat.com/browse/KEYCLOAK-14321 is fixed, so it would not have long time value to work on these files.

slaskawi (approved):

slaskawi:
Integrated, thanks @chlunde !

PR #244, opened 2020-08-31
KEYCLOAK-12023 Use Service Accounts

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-12023

## Additional Information

This Pull Request introduces using Keycloak Service Accounts for communication between Keycloak and the Operator. 

## Checklist:
- [x] Automated Tests


slaskawi:
@stianst @abstractj @vmuzikar @ASzc @miquelsi @pb82 @davidffrench @matskiv This PR contains an initial implementation for using Service Accounts for communicating with Keycloak.

When the Operator spins Keycloak up, it creates a "master" realm and attaches `<realm>-operator-cli` Client to it. Later on, the Operator connects to Keycloak using Client ID and Secret combination. If for some reason the Client is not there (e.g. initial bootstrap process) - Admin Credentials are used. 

There are still some open questions here:
- @stianst mentioned that we should use the Refresh Token to obtain a new Access Token every time we access Keycloak. We don't do that at the moment and it should probably be added in some other PR (this one already is quite large).
- The Operator doesn't create all CRs for all built-in clients. Since we create a CR for "master" realm, perhaps for symmetry reasons we should also bootstrap all built-in clients?
- We use quite weak Client authentication model - Client Credentials. At the moment, we don't support neither Signed JWTs nor mTLS. Is it fine to leave it with Client Credentials for now?

slaskawi:
Because of other team priorities, I'm putting this on hold.

abstractj:
@slaskawi I'd suggest to just close it. As I'm afraid this PR will get out of date. We can always reopen.

slaskawi:
Sure, let me close it.

slaskawi:
Answering open questions from the above:

> @stianst mentioned that we should use the Refresh Token to obtain a new Access Token every time we access Keycloak. We don't do that at the moment and it should probably be added in some other PR (this one already is quite large).

Unfortunately this is not the way it is done today (with or without this PR). This Pull Request doesn't change anything in this regard. I created [KEYCLOAK-17008](https://issues.redhat.com/browse/KEYCLOAK-17008) to introduce a cache for AccessTokens and introduce a Refresh Token mechanism.

> The Operator doesn't create all CRs for all built-in clients. Since we create a CR for "master" realm, perhaps for symmetry reasons we should also bootstrap all built-in clients?

After thinking about this for a while - I think it's unnecessary. We can always add this later.

> We use quite weak Client authentication model - Client Credentials. At the moment, we don't support neither Signed JWTs nor mTLS. Is it fine to leave it with Client Credentials for now?

Created [KEYCLOAK-15451](https://issues.redhat.com/browse/KEYCLOAK-15451) and [KEYCLOAK-15452](https://issues.redhat.com/browse/KEYCLOAK-15452). Again, the situation is the same - we didn't have the support for other clients types without this PR. This PR doesn't change anything here.

stianst:
Why is the operator creating a "master" realm? That bootstrapping is something that Keycloak already takes care of.

stianst:
We've fixed the issue with refresh tokens for service accounts, so there are no refresh tokens for service accounts anymore, and a new access token is just fetched with the clients credentials. That doesn't mean a client shouldn't re-use the access token during its availability though. With the whole re-consolidation loop I would imagine the operator does communicate quite a bit with KC, so it would be important to not fetch a new token every time it has to call KC admin APIs.

slaskawi:
> Why is the operator creating a "master" realm? That bootstrapping is something that Keycloak already takes care of.

@stianst This is required as `KeycloakClient` CRs use a `realmSelector` to find a Realm the Clients needs to be created in. 

> We've fixed the issue with refresh tokens for service accounts, so there are no refresh tokens for service accounts anymore, and a new access token is just fetched with the clients credentials. That doesn't mean a client shouldn't re-use the access token during its availability though. With the whole re-consolidation loop I would imagine the operator does communicate quite a bit with KC, so it would be important to not fetch a new token every time it has to call KC admin APIs.

@stianst Thanks for letting me know. I've updated [KEYCLOAK-17008](https://issues.redhat.com/browse/KEYCLOAK-17008).

slaskawi:
The problem with the CI is that after some time, the tokens got using Client Credentials grant start returning HTTP 403 Unauthorized. Oddly enough, after some time, this error disappears and the Operator starts to work properly again. Let me create a PR with the latest Keycloak build and then get back to this again.

slaskawi:
This Pull Request should be taken over by someone. @stianst @vmuzikar @abstractj any volunteers? If I don't hear anything back from you till the next week, I'll close it.

kungfuchicken:
This is highly valuable, and should be prioritized by official Redhat devs.

Kampe:
> This Pull Request should be taken over by someone. @stianst @vmuzikar @abstractj any volunteers? If I don't hear anything back from you till the next week, I'll close it.

Please don't close! This is HIGHLY valuable.

vmuzikar:
Hi @slaskawi! Thank you very much for this contribution. After thorough discussion with @stianst we eventually decided not to implement/merge this. We might reevaluate this in the future but for now it doesn't fit in the current operator roadmap. Thanks for understanding.

PR #247, opened 2020-09-15
KEYCLOAK-15476: Support external keycloak instances

jniebuhr:
## JIRA ID
KEYCLOAK-15476

## Additional Information
In it's current state, the operator only supports a keycloak that is created by the operator. In a setup where the keycloak is run outside of the cluster or possibly a multi-cluster setup, this keeps us from creating realms or clients on that keycloak. The same goes for externally created realms on that server.

## Checklist:
- [x] Automated Tests

## Additional Notes 
Documentation PR: https://github.com/keycloak/keycloak-documentation/pull/1047


deploy/crds/keycloak.org_keycloaks_crd.yaml
```diff
@@ -31,13 +31,25 @@ spec:
...
                 files that will be deployed in Keycloak.
               items:
                 type: string
               type: array
               x-kubernetes-list-type: set
+            external:
+              description: When set to true, this Keycloak will be marked external
+                and not be managed by this operator. It can then be used to create
+                realms, clients and users.
+              type: boolean
```
kfox1111, line 52: Another option may be to put credentialSecret and endpoint as properties of external. Then having the properties at all implies external.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -7,6 +7,10 @@ import (
 // KeycloakRealmSpec defines the desired state of KeycloakRealm.
 // +k8s:openapi-gen=true
 type KeycloakRealmSpec struct {
+	// When set to true, this KeycloakRealm will be marked external and not be managed by this operator.
+	// It can then be used to create clients and users.
+	// +optional
+	External bool `json:"external,omitempty"`
```
kfox1111, line 13: This is an interesting flag.... Could be made more generic. Would there ever be a case where you want to have keycloak manage a realm for a while, then stop but still be able to manage clients? If so, then maybe the flag should be:
Unmanaged bool
Managed external may make sense as well.

deploy/crds/keycloak.org_keycloaks_crd.yaml
jniebuhr: Sure, that would work. I would then add a `external.enabled` property as we're doing in `externalAccess.enabled`? I guess I would then have the same syntax in the `KeycloakRealm` CRD.

davidffrench:
I have not reviewed the PR itself, however, +1 from me on this variant.

I prefer it because at the end of the day it is all about user experience. Having a completely separate CR for using a non-operator created Keycloak is a little anti-intuitive in my opinion. Using the same CR with an override or in this case, the external field is much simpler to understand in my opinion. When a new user is looking at the available Keycloak CRs from the OLM interface would be initially confused with the similar but different CRs showing up and would need to look at the docs to understand what the difference is.

slaskawi:
@jniebuhr I believe we have all we need to move this forward. Could you please convert this to a final Pull Request and address all the comments we spoke about in the [Google Group](https://groups.google.com/g/keycloak-dev/c/ykKYfSY4cpw). Once you do that, please ping me and I'll review the Pull Request.

slaskawi:
@jniebuhr Do you have any plans to work on this in the near future? I'm considering reopening my Service Accounts PR (https://github.com/keycloak/keycloak-operator/pull/244), that needs to be rebased on top of this change. If you don't have enough time, I can take this over.

jniebuhr:
Hey @slaskawi, sorry for the delay. I will get to it this week.

slaskawi:
@jniebuhr Awesome! Once you finish, please ping me. I'll review the changes.

slaskawi:
@jniebuhr Any progress?

jniebuhr:
@slaskawi I've finalized the code changes. Working on the docs as we speak

jniebuhr:
I don't really get why CI is failing me, can someone restart it?

slaskawi:
> I don't really get why CI is failing me, can someone restart it?

I restarted the job. It seems there was some hiccup on Travis.

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
```diff
@@ -584,6 +584,11 @@ spec:
                     type: string
                 type: object
               type: array
+            unmanaged:
```
slaskawi, line 587: I think we could get away without using the `unmanaged` field. We could just scan `external.enabled`. If you agree with me, would you mind to remove it?

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -80,6 +87,15 @@ var (
 	PassthroughTLSTerminationType TLSTerminationType = "passthrough"
 )
 
+type KeycloakExternal struct {
+	// If set to true, this Keycloak will be treated as an external instance.
+	// The unmanaged field also needs to be set to true if this field is true.
+	Enabled bool `json:"enabled,omitempty"`
+	// The URL to use for the keycloak admin API. Needs to be set if external is true.
+	// +optional
```
slaskawi, line 95: This one is probably required. Otherwise it will fail in the runtime when trying to create a client.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -7,6 +7,10 @@ import (
 // KeycloakRealmSpec defines the desired state of KeycloakRealm.
 // +k8s:openapi-gen=true
 type KeycloakRealmSpec struct {
+	// When set to true, this KeycloakRealm will be marked as unmanaged and not be managed by this operator.
+	// It can then be used for targeting purposes.
+	// +optional
+	Unmanaged bool `json:"unmanaged,omitempty"`
```
slaskawi, line 13: If you agree with my earlier comment about removing `unmanaged` field, we could rename this to `external.enabled` for symmetry reasons with `Keycloak` CR.

slaskawi (changes requested):
I like the proposed solution a lot! There are just a few comments to be addressed.

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
kfox1111: I suggested it earler. External implies it will only ever be used for external things. I think it may be useful to have an unmanaged flag that is a little more general, as you can tell the operator, I'm going to manage this Realm manually, but I still may manage a few clients even when the whole keycloak is otherwise managed by the operator.

slaskawi: I see your comment [here](https://github.com/keycloak/keycloak-operator/pull/247#discussion_r488787797). 

The main reason I don't like `unmanaged` and `external.enabled` flags combination is because they result in exactly the same behavior - the Operator doesn't perform reconciliation and obtains Keycloak's URL from `external.url`. But you also have a point - if we add the `external.enabled` flag to the `KeycloakRealm` CR, it will look awkward. Using `managed` flag is much more descriptive.

I'm on the fence with this. But it probably won't hurt to add both flags for the sake of symmetry.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
slaskawi: Please disregard this, see https://github.com/keycloak/keycloak-operator/pull/247#discussion_r508239294

slaskawi:
@jniebuhr Could you please address my comment about the `external.url` (see [here](https://github.com/keycloak/keycloak-operator/pull/247#discussion_r507534389)) and rebase this PR? Then, please ping me. We should be ready to merge it.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
jniebuhr: Can we set this to required? I didn't find anything about conditional validation. If enabled is false, this field isn't required.

slaskawi: ooh I see what you mean. Yeah, good point! Let's leave it as is.

slaskawi:
@jniebuhr Since I have you here - would you mind to squash all those commits?

slaskawi:
Yes, that would be helpful. Thanks!

jniebuhr:
@slaskawi All checks passed, looks ready to be merged

slaskawi (approved):

slaskawi:
Integrated, thanks @jniebuhr !

PR #251, opened 2020-10-08
KEYCLOAK-14955 Support for running operator tests without admin permissions

vmuzikar:
## JIRA ID
KEYCLOAK-14955

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,11 @@ test/e2e-latest-image:
 	# Doesn't need cluster/prepare as it's done by operator-sdk. Uses a randomly generated namespace (instead of keycloak namespace) to support parallel test runs.
 	operator-sdk run local ./test/e2e --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count" --debug --verbose
 
+.PHONY: test/ibm-validation
+test/ibm-validation:
+	@echo Running the operator image in the cluster
+	operator-sdk test local ./test/e2e --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0" --operator-namespace $(NAMESPACE) --debug --verbose --global-manifest=deploy/empty-init.yaml --namespaced-manifest=deploy/operator.yaml
+
 .PHONY: test/e2e-local-image cluster/prepare setup/operator-sdk
 test/e2e-local-image: cluster/prepare setup/operator-sdk
 	@echo Running e2e tests with a fresh built operator image in the cluster:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -60,6 +60,11 @@ test/e2e-latest-image:
 	# Doesn't need cluster/prepare as it's done by operator-sdk. Uses a randomly generated namespace (instead of keycloak namespace) to support parallel test runs.
 	operator-sdk run local ./test/e2e --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count" --debug --verbose
 
+.PHONY: test/e2e-image-no-admin
```
slaskawi, line 63: I would prefer to replace one of our existing targets with this. The `test/e2e-latest-image` seems to be a good candidate.

README.md
```diff
@@ -133,6 +133,16 @@ Debug the e2e operator tests in [Goland](https://www.jetbrains.com/go/)
 | `make test/coverage/prepare` | Prepares coverage report from unit and e2e test results     |
 | `make test/coverage`         | Generates coverage report                                   |
 
+##### Running tests without cluster admin permissions
+It's possible to deploy CRDs, roles, role bindings, etc. separately from running the tests:
+1. Run `make cluster/prepare` as a cluster admin.
+2. Run `make test/e2e-image-no-admin` as a user. The user needs the following permissions to run te tests:
+```
+apiGroups: ["", "apps", "keycloak.org"]
+resources: ["persistentvolumeclaims", "deployments", "statefulsets", "keycloaks", "keycloakrealms", "keycloakusers", "keycloakclients", "keycloakbackups"]
```
slaskawi, line 142: Could you please tell me how this is used? Perhaps we don't need to mention it in the docs? Or maybe, it's better to write a `role.yaml` file dedicated to this use case?

slaskawi (changes requested):

Makefile
vmuzikar: It's basically a small hack to run the tests this way – we need to use custom manifests etc. I'm not sure we want to use it on regular basis (like `test/e2e-latest-image`). Also, you still need to manually run `make cluster/prepare` as an admin before this, so I don't really we should replace some existing target with this. This is really a special use case.

README.md
vmuzikar: This is just not forget what permissions are required to run the tests without cluster admin permissions. I'd say let's leave it to the cluster admins how will they use this – whether they create a new role or add a new rule to an existing role.

Makefile
slaskawi: Ok, understood. 

And where does it fail? What if we break it somewhere in the upstream? 

I would like to avoid a situation where we know about a potential failure late in the loop. If this configuration is needed, it should be tested in Travis in my opinion.

README.md
slaskawi: Ok, understood. Thanks!

vmuzikar:
@slaskawi I've renamed the make target and added a small disclaimer to the readme.

slaskawi (approved):

PR #253, opened 2020-10-13
KEYCLOAK-15895 "make test/e2e-local-image" runs the operator locally instead of deploying the image

vmuzikar:
## JIRA ID
KEYCLOAK-15895

## Checklist:
- [X] Automated Tests

## Additional Notes
Added also a small fix to make `setup/operator-sdk` work on macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,7 @@ test/e2e-local-image: setup/operator-sdk
 	docker build . -t keycloak-operator:test
 	@echo Running tests:
 	trap 'mv -f deploy/operator.yaml_bckp deploy/operator.yaml' EXIT; \
-	-operator-sdk test local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0" --image="keycloak-operator:test" --debug --verbose ./test/e2e
-	kubectl get pods --all-namespaces
+	operator-sdk test local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0" --image="keycloak-operator:test" --debug --verbose --up-local ./test/e2e
 
 .PHONY: test/coverage/prepare
 test/coverage/prepare:
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,14 @@
-# Other contants
+# Other constants
 NAMESPACE=keycloak
 PROJECT=keycloak-operator
 PKG=github.com/keycloak/keycloak-operator
 OPERATOR_SDK_VERSION=v0.18.2
-OPERATOR_SDK_DOWNLOAD_URL=https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu
+ifeq ($(shell uname),Darwin)
+  OPERATOR_SDK_ARCHITECTURE=x86_64-apple-darwin
+else
+  OPERATOR_SDK_ARCHITECTURE=x86_64-linux-gnu
+endif
+OPERATOR_SDK_DOWNLOAD_URL=https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-$(OPERATOR_SDK_ARCHITECTURE)
 MINIKUBE_DOWNLOAD_URL=https://github.com/kubernetes/minikube/releases/download/v1.9.2/minikube-linux-amd64
 KUBECTL_DOWNLOAD_URL=https://storage.googleapis.com/kubernetes-release/release/v1.18.0/bin/linux/amd64/kubectl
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -60,12 +65,16 @@ test/e2e-latest-image:
...
-.PHONY: test/e2e-local-image cluster/prepare setup/operator-sdk
-test/e2e-local-image: cluster/prepare setup/operator-sdk
+.PHONY: test/e2e-local-image
+test/e2e-local-image: setup/operator-sdk
 	@echo Running e2e tests with a fresh built operator image in the cluster:
 	docker build . -t keycloak-operator:test
+	@echo Backing up operator.yaml
+	@cp deploy/operator.yaml deploy/operator.yaml_bckp
+	@echo Modifying operator.yaml
+	@sed -i 's/imagePullPolicy: Always/imagePullPolicy: Never/g' deploy/operator.yaml
```
vmuzikar, line 75: The image is not uploaded to any Registry so it cannot be pulled.

Makefile
```diff
@@ -60,12 +65,16 @@ test/e2e-latest-image:
...
 	@echo Running tests:
-	operator-sdk test local --go-test-flags "-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0" --image="keycloak-operator:test" --namespace $(NAMESPACE) --up-local --debug --verbose ./test/e2e
+	bash -c "trap 'mv -f deploy/operator.yaml_bckp deploy/operator.yaml' EXIT; operator-sdk test local --go-test-flags '-tags=integration -coverpkg ./... -coverprofile cover-e2e.coverprofile -covermode=count -timeout 0' --image='keycloak-operator:test' --debug --verbose ./test/e2e"
```
slaskawi, line 77: I would split it into two steps for readability.

If you want to ignore `mv` exit code, just use `mv <...> || true` as we did a couple of times in this file.

slaskawi (commented):
I would suggest small changes but let me assign this to @miquelsi

Makefile
vmuzikar: I needed to simulate a `finally` block. ;) I don't need to ignore the `mv` exit code. What I need is to perform the clean up regardless the test result (`operator-sdk` exit code) and at the same time not ignore the test result. We want the make target to fail if tests fail. The problem here is that make won't proceed with a command if the previous command failed. Another problem was that make uses a separate shell to execute each command, so `trap` wouldn't work if split into two steps.
If researched multiple solutions and this seems like the most straightforward/simple one.

slaskawi: Ok, understood! Thanks for the explanation!

Makefile
miquelsi: Should we also change the image to **keycloak-operator:test** in **deploy/operator.yaml**?

vmuzikar: No, not permanently. By default it should use current master.

It's overridden only during testing and then reverted back to master.

vmuzikar: This is done transparently by Operator SDK's `--image` parameter.

vmuzikar: You can double check it in the test log – it's printed out from where the operator is deployed.

miquelsi: All right. Then I have no further objections. Thanks @slaskawi!

miquelsi (approved):

vmuzikar:
Please wait with the merge! I just noticed Travis failed.

slaskawi:
@vmuzikar This now requires a rebase.

vmuzikar:
@slaskawi Hmmm, I'm currently getting the following error in Travis:
```
Step 1/8 : FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS build-env
326received unexpected HTTP status: 500 Internal Server Error
```
Any idea where to report it if it makes sense, or do we just wait?

slaskawi:
Ok, so the CI errors in this CI doesn't seem to be connected to OpenShift registry failures.

vmuzikar:
I'm closing this for now as I can't invest any more time into this at the moment. The problem there is that the operator deployed in the cluster cannot access the keycloak service – it is unable to resolve keycloak hostname. It seems like some weird DNS problem which is present in minikube probably in combination with Travis environment.

I'll try to investigate this later (probably a few days).

slaskawi:
I think I have some idea...

It seems the Operator Pod doesn't inherit the DNS configuration from
Kubernetes. I'm not sure why this happens but I may also want to run a few
debug runs to wrap my head around this.

But let's dig into this at the end of the sprint. Now we have a few things
with higher priority.

vmuzikar:
As my mind was full of this, I just reallized something. :) To access the KC service, the operator is using a hostname (keycloak.local) that is provided by ingress. But! Ingress is meant to be accessed from outside, not internally from the cluster. Am I correct?

Why it works for me locally is because on macOS both K8s from Docker for Mac and VBox as well use the same network as the host. That means the KC service is accessible through the external hostname provided by ingress. That is not the case for Linux and that's why it doesn't work in Travis.

slaskawi:
@vmuzikar Do you plan to work on this anytime soon?

vmuzikar:
@slaskawi Yes, I can work on this (to finish what I started). :)

vmuzikar:
jFTR – it is really a bug. Created https://issues.redhat.com/browse/KEYCLOAK-16220.

slaskawi:
@vmuzikar Ok, are you actively working on this? If not, can we close it for now? I'd like to know how many active PRs we have.

vmuzikar:
Yes, we can close this. We need to fix KEYCLOAK-16220 first and then implement changes from this to GH Actions.

PR #269, opened 2020-11-17
KEYCLOAK-14683 fix status loop

mathianasj:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-14683

## Additional Information
Only update status if a change was made

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/controller/keycloak/keycloak_controller.go
+++ b/pkg/controller/keycloak/keycloak_controller.go
@@ -3,6 +3,7 @@ package keycloak
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/keycloak/keycloak-operator/version"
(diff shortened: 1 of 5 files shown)


slaskawi (changes requested):
Unfortunately it seems you scratched the surface and make a bigger problem visible by this Pull Request.

Starting from Day 1 of our Keycloak Operator implementation we were always sending the updated objects to the Operator SDK. We very heavily relied on the Client Cache behavior that Operator SDK sends only modified objects from its cache to the Kubernetes API Server. The main reasoning behind our approach was to simplify the Operator's logic. It seems our assumption no longer holds and I see all updates are always sent to Kubernetes (you can spot this by increasing number in `metadata.resourceVersion` for all our objects).

As this is being part of a larger problem, we need a better solution. There are a few options that I can think of:
- Build cache on Keycloak Operator side. It's the most flexible option but also the hardest to get right.
- Try to use `r.client.Status().Patch` and experiment with a few `PatchOptions`. Perhaps we can easily translate our code to get the behavior we want. 
- Find some other way to achieve this.

@pb82 You might be interested in this.

slaskawi:
@mathianasj Any update on this one?

mathianasj:
@slaskawi I'll see if I have some time this week to take a look if we can update that way

slaskawi:
Any update, @mathianasj ?

slaskawi:
@mathianasj Any update?

slaskawi:
Let me close this for now. Please reopen when ready, @mathianasj

PR #272, opened 2020-11-19
[KEYCLOAK-16368] CI is broken due to some changes on GitHub actions

abstractj:
## [KEYCLOAK-16368](https://issues.redhat.com/browse/KEYCLOAK-16368)

Replaces the usage of [medyagh/setup-minikube@master](https://github.com/medyagh/setup-minikube)

## Additional Information

Unfortunately, the GitHub actions for Minikube still use `add-path` which is deprecated by GitHub and no longer functional. We tried to submit [a pull-request](https://github.com/medyagh/setup-minikube/pull/13) to the author, but we didn't hear anything back until now.

Considering that this is critical, I'd suggest to include the changes in this PR.

## Additional Notes 

* `ACTIONS_ALLOW_UNSECURE_COMMANDS` solves the issue, but it is unsafe.
* Coveralls was intentionally removed as it was already broken a while ago. We may revisit it in the future, but not a priority for now.
* Supersedes #271

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,30 +18,15 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
     - name: Unit tests
-      env:
-        COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        make test/unit test/goveralls
-    - name: Start minikube
-      id: minikube
-      env:
-        CHANGE_MINIKUBE_NONE_USER: true
-        MINIKUBE_WANTUPDATENOTIFICATION: false
-        MINIKUBE_WANTREPORTERRORPROMPT: false
-      uses: medyagh/setup-minikube@master
-    - name: Setup minikube
-      run: |
-        sudo ./hack/modify_etc_hosts.sh "keycloak.local"
-        minikube addons enable ingress
+        make test/unit
     - name: Run e2e tests
-      env:
-        COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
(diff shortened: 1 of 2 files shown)


slaskawi (approved):

miquelsi (approved):

Makefile
```diff
@@ -157,8 +157,8 @@ code/lint:
 ##############################
 # CI                         #
 ##############################
-.PHONY: setup/travis
-setup/travis:
+.PHONY: setup/github
+setup/github:
```
ASzc, line 161: Technically these targets should be left as PHONY, because they don't produce a file with the target name. Doesn't really matter though.

.github/workflows/ci.yml
```diff
@@ -18,30 +18,15 @@ jobs:
...
-    - name: Setup minikube
-      run: |
-        sudo ./hack/modify_etc_hosts.sh "keycloak.local"
-        minikube addons enable ingress
+        make test/unit
     - name: Run e2e tests
-      env:
-        COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
+        make setup/github
```
ASzc, line 25: So, the idea is to replace the github action with this makefile target? i.e. option #2 in the JIRA?

Makefile
ASzc: Oops, I misread the diff, nevermind :)

.github/workflows/ci.yml
ASzc: I guess my thought is: why weren't we just doing that before?

slaskawi:
Integrated, thanks @abstractj !

PR #243, opened 2020-08-17
INTLY-8357 Pod monitor removal

obrienrobert:
## JIRA ID
https://issues.redhat.com/browse/INTLY-8357

## Additional Information
This PR removes the PodMonitor resource which was added as part of the following PR: https://github.com/keycloak/keycloak-operator/pull/114.

## Verification Steps

- Checkout this branch
- Apply all the CRDs etc: `make cluster/prepare`
- Deploy the operator locally: `operator-sdk run --local --namespace "keycloak"`
- Apply the Keycloak CR: `oc apply -f deploy/examples/keycloak/keycloak.yaml`
- Apply the Keycloak Realm CR: `oc apply -f deploy/examples/realm/basic_realm.yaml`
- Verify that no PodMonitor has been created and that the operator logs contain **no** reference to a PodMonitor resource:

```
{"level":"info","ts":1597657288.0097225,"logger":"controller_keycloak","msg":"Reconciling Keycloak","Request.Namespace":"keycloak","Request.Name":"example-keycloak"}
{"level":"info","ts":1597657288.0260136,"logger":"action_runner","msg":"(    0)    SUCCESS Update Keycloak admin secret"}
{"level":"info","ts":1597657288.0463824,"logger":"action_runner","msg":"(    1)    SUCCESS update keycloak prometheus rule"}
{"level":"info","ts":1597657288.0647643,"logger":"action_runner","msg":"(    2)    SUCCESS update keycloak service monitor"}
{"level":"info","ts":1597657288.1027825,"logger":"action_runner","msg":"(    3)    SUCCESS update keycloak grafana dashboard"}
{"level":"info","ts":1597657288.1177003,"logger":"action_runner","msg":"(    4)    SUCCESS Update Database Secret"}
{"level":"info","ts":1597657288.1337612,"logger":"action_runner","msg":"(    5)    SUCCESS Update Postgresql PersistentVolumeClaim"}
{"level":"info","ts":1597657288.1505246,"logger":"action_runner","msg":"(    6)    SUCCESS Update Postgresql Deployment"}
{"level":"info","ts":1597657288.166885,"logger":"action_runner","msg":"(    7)    SUCCESS Update Postgresql KeycloakService"}
{"level":"info","ts":1597657288.18427,"logger":"action_runner","msg":"(    8)    SUCCESS Update keycloak Service"}
{"level":"info","ts":1597657288.2011492,"logger":"action_runner","msg":"(    9)    SUCCESS Update keycloak Discovery Service"}
[...]

PR #274, opened 2020-11-28
Add list podmonitors to role

codejamninja:
The operator needs to have permission to list podmonitors for the monitoring.coreos.com api.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -72,6 +72,7 @@ rules:
   resources:
   - servicemonitors
   - prometheusrules
+  - podmonitors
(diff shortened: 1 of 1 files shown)


slaskawi:
Thanks for the Pull Request @codejamninja !

I believe this one should not be required as Pod Monitors were removed in https://github.com/keycloak/keycloak-operator/pull/243. The change will be available in Keycloak Operator 12.

Please reopen this Pull Request if this doesn't solve your issue.

codejamninja:
Ok, thanks @slaskawi

PR #280, opened 2020-12-02
KEYCLOAK-16523 Respect proxy settings

nbjohnson:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16523

## Additional Information
Update httpTransport to maintain proxy settings for http requests.

## Checklist:
- [x] Automated Tests

### DIFF
--- a/pkg/common/client.go
+++ b/pkg/common/client.go
@@ -726,9 +726,9 @@ func (c *Client) login(user, pass string) error {
 
 // defaultRequester returns a default client for requesting http endpoints
 func defaultRequester() Requester {
-	transport := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // nolint
-	}
+	transport := http.DefaultTransport.(*http.Transport).Clone() // nolint
+	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true} // nolint
+
 	c := &http.Client{Transport: transport, Timeout: time.Second * 10}
 	return c
 }


slaskawi (approved):
LGTM, @abstractj @vmuzikar @miquelsi @ASzc Would you also like to review this?

vmuzikar:
I think some test is missing here. WDYT @slaskawi?

slaskawi:
That's a good point, would you mind adding it @nbjohnson ?

nbjohnson:
Thanks for taking the time to take a look at my PR. What kind of test are you looking for? A more automated go test? Or more of a by hand here is an environment situation, in this case proxy situation, and see it doesn't work properly with the proxy enabled and this patch fixes it? Just let me know what you are looking for and I will get right on it. Again, Thanks!

slaskawi:
Just a simple unit test is absolutely fine. We just want to make sure we
won't break this in the future.

slaskawi:
@nbjohnson Any update on that missing test?

nbjohnson:
I apologize for the delay, got busy with other things. Yes I am working on this, just having some difficulty finding the best way to make a unit test for this edge case. The tests don't seem to provide adequate access to the requester object to check the status of the proxy setting, so working on figuring that out.

nbjohnson:
So as a follow-up I have continued looking through the test setup and I am not seeing a good way to be able to integrate this into the tests. The problem here is that it is buried in the client file with no real good way to be able to trigger this line of code as well as being able to see the settings of the transport. From what I understand this code is only run with the AuthenticatedClient function, which is triggered for example when a new realm is created and the operator tries to negotiate with the front end.

The main issue was that http.Transport does not pick up proxy settings, whereas http.DefaultTransport does pick up proxy settings. In running that code in a debugger I can see that my change allows the transport to populate with proxy environment variables where it was nil previously. In running locally I can spin up an instance and add a realm and client without issue.

Your main concern was making sure this won't be broken in the future, but I am not sure of a good way to have a test for that, I am open to suggestions though, other than not allowing http.Transport to be used in the codebase and only http.DefaultTransport which will always pick up environment proxy settings. Some of your tests use a test server and set up calls using http.DefaultTransport so your tests had this correct, but the code in this specific case was not. Looking back at the revision history it looks like this error was in here since the initial commit.

slaskawi:
Fair enough, let's get this in.

slaskawi:
Integrated, thanks @nbjohnson !

nbjohnson:
Ok I see it merged in and it fixed the issue we were having. Thanks for working with me to get this in and your quick responses.

PR #285, opened 2020-12-09
KEYCLOAK-16359 Add realm email settings to realm CRD

dhx-teuto:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16359

This pull request is only intended for the last commit, the others are minor fixes to code generation etc. that should be done beforehand.
I will rebase as needed.

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
@@ -134,6 +134,10 @@ type KeycloakAPIRealm struct {
 	// Failure Reset Time
 	// +optional
 	MaxDeltaTimeSeconds *int32 `json:"maxDeltaTimeSeconds,omitempty"`
+
+	// Email
+	// +optional
+	SMTPServer map[string]string `json:"smtpServer,omitempty"`
 }
 
 // https://www.keycloak.org/docs-api/10.0/rest-api/index.html#_userfederationproviderrepresentation
--- a/test/e2e/keycloak_realm_test.go
+++ b/test/e2e/keycloak_realm_test.go
@@ -70,6 +70,15 @@ func getKeycloakRealmCR(namespace string) *keycloakv1alpha1.KeycloakRealm {
 				MinimumQuickLoginWaitSeconds: &[]int32{60}[0],
 				MaxFailureWaitSeconds:        &[]int32{900}[0],
 				MaxDeltaTimeSeconds:          &[]int32{43200}[0],
+				SMTPServer: map[string]string{
+					"starttls":        "",
+					"auth":            "",
+					"host":            "smtp.server",
+					"from":            "sso@example.com",
+					"fromDisplayName": "Example Company",
+					"envelopeFrom":    "sso@example.com",
+					"ssl":             "",
+				},
 			},
 		},
 	}
--- a/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
@@ -475,6 +475,11 @@ spec:
                 resetPasswordAllowed:
                   description: Forgot password
                   type: boolean
+                smtpServer:
+                  additionalProperties:
+                    type: string
+                  description: Email
+                  type: object
                 sslRequired:
                   description: Require SSL
                   type: string
--- a/pkg/apis/keycloak/v1alpha1/zz_generated.deepcopy.go
+++ b/pkg/apis/keycloak/v1alpha1/zz_generated.deepcopy.go
@@ -502,6 +502,13 @@ func (in *KeycloakAPIRealm) DeepCopyInto(out *KeycloakAPIRealm) {
 		*out = new(int32)
 		**out = **in
 	}
+	if in.SMTPServer != nil {
+		in, out := &in.SMTPServer, &out.SMTPServer
+		*out = make(map[string]string, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
+		}
+	}
 	return
 }
 


slaskawi:
I'll hold off with comments until we get your previous PRs in. Then, I'll ask you for a rebase and we'll take it from there.

dhx-teuto:
@slaskawi rebased.

slaskawi:
I've re-triggered the CI to pick the latest changes up.

@vmuzikar May I kindly ask you to review this one?

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -109,6 +109,10 @@ type KeycloakAPIRealm struct {
 	// Require SSL
 	// +optional
 	SslRequired string `json:"sslRequired,omitempty"`
+
+	// Email
```
vmuzikar, line 113: Just a nitpick. Maybe "SMTP server" would be more descriptive here?

vmuzikar (commented):
@dhx-teuto Thank you for your contribution! Added one comment, other than that looks good.

vmuzikar (changes requested):
I overlooked the linting error. Could you please fix it?

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -109,6 +109,10 @@ type KeycloakAPIRealm struct {
...
+	// +optional
+	SmtpServer map[string]string `json:"smtpServer,omitempty"`
```
apapia, line 115: Would it not be more user friendly to use a struct with specific fields rather than a map here?  This part of the API is not well documented.  I had to use the admin console and inspect the page to see what parameters were being passed.  I would be nice to have a struct so that it is obvious how to configure this.

vmuzikar: This crossed my mind too. However, KC [uses a plain map](https://github.com/keycloak/keycloak/blob/f423c0dc5138fe14b985c17b46ff433da4056d25/core/src/main/java/org/keycloak/representations/idm/RealmRepresentation.java#L151) for this too. So it's a question whether we want to keep it consistent with the KC implementation, or improve it at the Operator side.

dhx-teuto: I totally agree with both of you. My impression was that 1:1 was the main goal https://issues.redhat.com/browse/KEYCLOAK-14321 therefore I used the plain map as well.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
dhx-teuto: All settings of course relate to the SMTP server in one way or another, however since it also includes things like From and display name, I'd like to stick to the more general Admin UI labelling `Email`.

dhx-teuto:
Fixed `SmtpServer` as per linter suggestion, rebased.

vmuzikar (approved):

slaskawi:
@dhx-teuto Would it be possible to add this into some of the existing integration tests?

test/e2e/keycloak_realm_test.go
```diff
@@ -62,6 +62,15 @@ func getKeycloakRealmCR(namespace string) *keycloakv1alpha1.KeycloakRealm {
...
 				DisplayName: "Operator Testing Realm",
+				SMTPServer: map[string]string{
+					"starttls":        "",
+					"auth":            "",
+					"host":            "smtp.server",
+					"from":            "sso@example.com",
+					"fromDisplayName": "Example Company",
+					"envelopeFrom":    "sso@example.com",
+					"ssl":             "",
+				},
```
dhx-teuto, line 73: @slaskawi sure

dhx-teuto:
@vmuzikar this one should be ready to merge

slaskawi (approved):

slaskawi:
Integrated, thanks @dhx-teuto

KleyLima (commented):
I am trying to use this feature in operator and I wanna know which and/or where is defined the allowed fields for `SMTPServer`. Thanks in advance.

dhoextermann:
The easiest way to get a configuration is to just copy and paste from a realm export.
Otherwise have a look at the e2e test https://github.com/keycloak/keycloak-operator
https://github.com/keycloak/keycloak-operator/blob/f2d0370290d6abe91724b2536748aa6b1245d1e1/test/e2e/keycloak_realm_test.go#L73-L81

KleyLima:
Thanks for the fast reply, I saw that test, but in comparison with the Keycloak's options in the screen to configure Email looks like it's missing some fields like port, login, and password (when the `auth` flag is trigged).

dhoextermann:
You're right, the export I used for that test did not include all possible options.
As I said, the easiest way to obtain a snippet is probably just configuring it in the GUI and exporting it.

[It was remarked in this PR that a struct would be more user friendly](https://github.com/keycloak/keycloak-operator/pull/285#discussion_r554099867), however the operator tries to stay as close to the Keycloak datastructures as possible and there it's unfortunately just a `Map<String, String>`.

If one person comes here there might be more, so here's a snippet with more fields
```yaml
smtpServer:
  auth: 'true'
  envelopeFrom: envelopefrom@example.com
  from: from@example.com
  fromDisplayName: From Display
  host: smtp.example.com
  password: password
  port: '25'
  replyTo: replyto@example.com
  replyToDisplayName: Reply To Display
  ssl: 'true'
  starttls: 'true'
  user: username
```
The docs don't go into that much detail (yet) but if you think it might be helpful feel free to create a PR.

KleyLima:
Thanks for the explanation!

KleyLima:
I am trying to put this settings in a yaml to take changes in kubernetes, I see in the issue of this feature `Fix Version: 13.0.0` so this feature isn't available? I an using keycloak-operator 12.0.3, and edited the realm CRD to comport this change, just link in the changes of the feature. The configuration don't take effects, something I do wrong?

dhoextermann:
Changing the CRD does not help. The values have to pass through the (Go) types that are compiled into the operator.

KleyLima:
I do the same mod that he does in the  `pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go` adding the type smtpSender, thats is what you are talking about?

dhoextermann:
yes, but

1. it's `smtpServer`, make sure the spelling is correct
2. keep in mind realms don't get updated once they're created so it has to be present in the first version of the realm CR you create

KleyLima:
Spelling is correct, and I am exactly trying to add this config to an existing realm, I will create a new with this yaml and see if the config takes place, thanks

hobbypunk90:
i have a little question, is there any option to get the smtp auth password from a secret?

hxtmdev:
> i have a little question, is there any option to get the smtp auth password from a secret?

Not that I remember from when I implemented it [because the data structures were supposed to mirror the Keycloak data structure as closely as possible](https://github.com/keycloak/keycloak-operator/pull/285#discussion_r555722583). But I am currently not up to date with the newest developments and the Jira issue has since been closed.

I suggest you try the [Forum](https://keycloak.discourse.group/) and the [mailing list](https://groups.google.com/forum/#!forum/keycloak-user) or peek into the code.

PR #282, opened 2020-12-09
[KEYCLOAK-16611] Add listType to realm/RedirectorIdentityProviderOverride

dhx-teuto:
https://issues.redhat.com/browse/KEYCLOAK-16611

This prevents a warning while running make code/gen.

## Additional Information
https://github.com/kubernetes/kube-openapi/blob/master/pkg/idl/listtype_test.go

PR #287, opened 2020-12-09
Fix code/gen generated files

dhx-teuto:
A previous commit seems to not have run an unmodified make code/gen.


slaskawi:
This will probably get automatically fixed once we get https://github.com/keycloak/keycloak-operator/pull/282 in. Closing.

PR #288, opened 2020-12-09
KEYCLOAK-16362 Add roles and client roles to realm/client CRDs

dhx-teuto:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16362

This pull request is only intended for the last commit, the others are minor fixes to code generation etc. that should be done beforehand.
I will rebase as needed.

## Checklist:
- [x] Automated Tests


dhx-teuto:
I believe my Jira issue might be a duplicate of https://issues.redhat.com/browse/KEYCLOAK-13370
Please let me know if you want the commit message changed.

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
```diff
@@ -13,6 +13,11 @@ type KeycloakClientSpec struct {
 	// Keycloak Client REST object.
 	// +kubebuilder:validation:Required
 	Client *KeycloakAPIClient `json:"client"`
+	// Client Roles
+	// +optional
+	// +listType=map
```
slaskawi, line 18: Could we use an atomic list here?

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
```diff
@@ -93,7 +98,7 @@ type KeycloakAPIClient struct {
 	Attributes map[string]string `json:"attributes,omitempty"`
 	// True if Full Scope is allowed.
 	// +optional
-	FullScopeAllowed bool `json:"fullScopeAllowed,omitempty"`
+	FullScopeAllowed *bool `json:"fullScopeAllowed,omitempty"`
```
slaskawi, line 101: This change should probably be excluded from this PR due to https://github.com/keycloak/keycloak-operator/pull/282

test/e2e/keycloak_client_test.go
```diff
@@ -104,3 +109,67 @@ func keycloakClientBasicTest(t *testing.T, framework *test.Framework, ctx *test.
 func externalKeycloakClientBasicTest(t *testing.T, framework *test.Framework, ctx *test.Context, namespace string) error {
 	return WaitForClientToBeReady(t, framework, namespace, testExternalKeycloakClientCRName)
 }
+
+func getKeycloakClientWithRolesCR(namespace string, roleNames []string) *keycloakv1alpha1.KeycloakClient {
```
slaskawi, line 113: @vmuzikar Would you mind to have a look at this bit?

slaskawi (changes requested):
Could you also squash your commits?

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
dhx-teuto: My reasoning was that the reconciler code I wrote treats it like map: It identifies roles by their names when figuring out what to create/update/delete and just assumes there aren't any duplicates (which the map type would ensure). The mergeability is just a nice side effect.

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
dhx-teuto: The commit is merely included in the PR to make the PR testable on it's own. As written in the PR description I'll rebase when that commit is in master.

vmuzikar:
@dhx-teuto Could you please rebase to resolve the conflict? Thanks.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -108,6 +110,68 @@ type KeycloakAPIRealm struct {
 	// Require SSL
 	// +optional
 	SslRequired string `json:"sslRequired,omitempty"`
+
+	// Roles
+	// +optional
+	Roles *RolesRepresentation `json:"roles,omitempty"`
```
vmuzikar, line 116: *A generic comment, unrelated to this specific line.*

@slaskawi We currently don't support updating realms through the operator. Correct? This means users won't be able to manage realm roles after the realm was created, even though roles are not part of the "closest" realm setting in KC. Are we happy with this? Shouldn't we rather have realm roles as an independent CRD to better reflect KC's behaviour?

test/e2e/keycloak_client_test.go
```diff
@@ -104,3 +109,67 @@ func keycloakClientBasicTest(t *testing.T, framework *test.Framework, ctx *test.
...
+	err = GetNamespacedObject(framework, namespace, testKeycloakClientCRName, &client)
+	if err != nil {
+		return err
+	}
+	client.Spec.Roles = getKeycloakClientRoles([]string{"b", "c"})
+	err = Update(framework, &client)
+	if err != nil {
+		return err
+	}
+	err = ensureOperatorRan(t, framework, ctx, namespace)
```
vmuzikar, line 152: I'm not sure I'm happy with this approach. It feels to me a bit nasty. Could we rather somehow check if the roles were truly updated in the CR?

vmuzikar (changes requested):
@dhx-teuto Thank you for your contribution! Added a few comments + a few subjective nit picks (fix those only if you want ;)).

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -108,6 +110,68 @@ type KeycloakAPIRealm struct {
...
+
+// https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_rolesrepresentation
+type RolesRepresentation struct {
+	// Client Roles
+	// +optional
+	Client map[string]RoleRepresentationArray `json:"client,omitempty"`
+
+	// Realm Roles
+	// +optional
+	Realm []RoleRepresentation `json:"realm,omitempty"`
```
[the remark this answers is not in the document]
dhx-teuto: The reasoning was to stick with `[]RoleRepresentation`, `RoleRepresentationArray` was just a workaround to satisfy `make code/gen`. But I can change it if you want.

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
```diff
@@ -13,6 +13,11 @@ type KeycloakClientSpec struct {
 	// Keycloak Client REST object.
 	// +kubebuilder:validation:Required
 	Client *KeycloakAPIClient `json:"client"`
+	// Client Roles
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	Roles []RoleRepresentation `json:"roles,omitempty"`
```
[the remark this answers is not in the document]
dhx-teuto: The reasoning was to stick with `[]RoleRepresentation`, `RoleRepresentationArray` was just a workaround to satisfy `make code/gen`. But I can change it if you want.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
```diff
@@ -43,9 +43,43 @@ func (i *KeycloakClientReconciler) Reconcile(state *common.ClientState, cr *kc.K
...
+
+	rolesDeleted := roleDifference(state.Roles, cr.Spec.Roles)
+	for _, role := range rolesDeleted {
+		desired.AddAction(i.getDeletedClientRoleState(state, cr, &role))
+	}
+
+	rolesUpdated := roleDifference(roleDifference(cr.Spec.Roles, rolesNew), rolesDeleted)
+	for _, role := range rolesUpdated {
+		desired.AddAction(i.getUpdatedClientRoleState(state, cr, &role))
+	}
```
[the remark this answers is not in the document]
dhx-teuto: Yes it does an update on every change. IIRC there are efforts to not run all the time for unchanged resources?!

pkg/controller/keycloakclient/keycloakclient_reconciler.go
```diff
@@ -43,9 +43,43 @@ func (i *KeycloakClientReconciler) Reconcile(state *common.ClientState, cr *kc.K
...
+}
+
+func roleDifference(a []kc.RoleRepresentation, b []kc.RoleRepresentation) (d []kc.RoleRepresentation) {
+	remove := make(map[string]bool)
+	for _, role := range b {
+		remove[role.Name] = true
```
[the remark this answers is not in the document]
dhx-teuto: Not being able to rename a role is a good argument if that should be supported by the operator in this way.
The name fallback is something I proposed elsewhere as well https://issues.redhat.com/browse/KEYCLOAK-16624 but in this case I think it would have confusing side effects as the ID in the resource (although respected if present) is always set to the resulting ID on creation https://github.com/keycloak/keycloak-operator/blob/2c058104137c861a1c64d721489d1bc3a1386cb1/pkg/common/cluster_actions.go#L133
Or am I confusing different kinds of IDs here?
I think the difference between renaming and recreating roles would be that the latter implicitly removes all associations with the role, correct?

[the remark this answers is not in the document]
vmuzikar: > I think the difference between renaming and recreating roles would be that the latter implicitly removes all associations with the role, correct?

Good point. I didn't realize this.
So I think we really need to rely on the IDs with the name as a fallback. The fallback could be used only in the `roleDifference` function. The name would never be set as the ID if that's what you meant that would be confusing.

The question is whether we really support roles renaming. The UI (Admin Console) doesn't support this but the REST API apparently does. So just to be on the safe side, I think operator should be ready for it.

test/e2e/keycloak_client_test.go
dhx-teuto: The closest I got was looking into `managedFields` with `manager: keycloak-operator` but while it seems to be a better approach to look in the CR than to "implement the barrier" in this way I did not find information on how to use this in a reliable and clear way.
The alternative of manually setting status to something the operator will change again did not work either since `.status` cannot simply be updated.
The other tests just wait for the status but for testing a true update this is simply not enough.
It is a workaround but unfortunately the most reliable and clear way I could find so far that does not include accessing the KC API directly from the test project (then it could just check for the expected changes).

vmuzikar: AFAIK, it should be possible to retrieve current CR. Please see [`GetNamespacedObject`](https://github.com/keycloak/keycloak-operator/blob/e655f80a246509ff13bd4d607a40fc77bd2a004f/test/e2e/utils.go#L213) usages.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -108,6 +110,68 @@ type KeycloakAPIRealm struct {
...
+}
+
+// https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_rolerepresentation
+type RoleRepresentation struct {
+	// Role Attributes
+	// +optional
+	Attributes map[string]string `json:"attributes,omitempty"`
```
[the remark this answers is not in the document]
dhx-teuto: Thanks for pointing that out, I'll replace it with a `StringArray` workaround like `RoleRepresentationArray` to satisfy `make code/gen`.

test/e2e/keycloak_client_test.go
dhx-teuto: Yes that is undoubtedly possible. What approach are you suggesting? Do you have information that would suggest the `managedFields` approach is reliable and portable enough to not break from slight changes/updates to the e2e test environment?

vmuzikar: I'd just fetch a fresh client object using `GetNamespacedObject` and check the roles are really there. That should be robust enough. WDYT?

dhx-teuto: The roles are in the k8s object the instant `Update()` (L148) completes.
The test should ensure the operator was actually able to perform the necessary changes on the KC server and I think this can only be verified by either checking on the KC server (which is not done in any other e2e test AFAICS) or checking for success in the CR after the operator tried to reconcile (which is what's happening right now).
The tricky part is making sure the operator actually tried to reconcile and we're not just going with the previous `phase: Reconciling` because the operator was too slow / e2e tests too fast.
Other e2e tests don't have this problem as they don't try to update from a successfully reconciled state.

pkg/common/client.go
```diff
@@ -753,6 +786,10 @@ type KeycloakInterface interface {
 	UpdateClient(specClient *v1alpha1.KeycloakAPIClient, realmName string) error
 	DeleteClient(clientID, realmName string) error
 	ListClients(realmName string) ([]*v1alpha1.KeycloakAPIClient, error)
+	ListClientRoles(clientID, realmName string) ([]v1alpha1.RoleRepresentation, error)
+	CreateClientRole(specClient *v1alpha1.KeycloakAPIClient, role *v1alpha1.RoleRepresentation, realmName string) (string, error)
+	UpdateClientRole(specClient *v1alpha1.KeycloakAPIClient, role *v1alpha1.RoleRepresentation, realmName string) error
+	DeleteClientRole(specClient, role, realmName string) error
```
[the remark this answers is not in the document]
dhx-teuto: Definitely, thanks for pointing that out.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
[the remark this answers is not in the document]
dhx-teuto: Nevermind, with a `[]string` no additional type is necessary https://github.com/kubernetes-sigs/controller-tools/issues/213#issuecomment-507844871
Fixed as per your suggestion.

dhx-teuto:
Pushed some of the requested changes, will work on the rest + rebase later.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
dhx-teuto: While I appreciate having the functionality to rename roles I think this makes things rather intransparent as matching with fallback has some ambiguities here. How for example should the operator even behave when reading
```
role(id1, name1)
```
from the server and getting
```
role(id1)
role(name1)
```
from the CR?
Furthermore, in the spirit of having easily readable CRs I would prefer being able to create and update roles with just a name and the non-default fields set, so I don't think forcing IDs would be the best solution either.
I'd like to keep things as they are now (matching by name), though this does of course sacrifice the (not available to the UI) possibility to rename roles while keeping all associations (which would get lost on re-create).
Any other opinions on this?

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
vmuzikar: @slaskawi Could you please confirm this is intended behaviour?

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
vmuzikar: I would still like to support role renaming, even though it might be a corner case. IMHO since it is supported at the API side so we should support it in the operator as well.

How about this.
* Use `role.Id` instead of `role.Name` in `roleDifference`.
* Some new roles already have ID set. And if some role in the CR doesn't have ID, take it as a new role.

WDYT?

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
vmuzikar: @slaskawi Could you please comment on this?

test/e2e/keycloak_client_test.go
vmuzikar: Ok, I understand. I think `ensureOperatorRan` should be at least placed into more generic [`utils`](https://github.com/keycloak/keycloak-operator/blob/e655f80a246509ff13bd4d607a40fc77bd2a004f/test/e2e/utils.go) in case some other test needs it in the future.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
dhx-teuto: I'll rewrite the whole reconciliation function to make renames possible. It will have further implications like specifying a non-matching ID leading to re-creation. I will continue this later.

test/e2e/keycloak_client_test.go
dhx-teuto: Turns out directly asking the KC server is easier than expected. It also seems cleaner. See updated commit.

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
slaskawi: I believe we need a rebase here.

slaskawi:
> I believe my Jira issue might be a duplicate of https://issues.redhat.com/browse/KEYCLOAK-13370
Please let me know if you want the commit message changed.

@dhx-teuto  That's fine. I've just closed https://issues.redhat.com/browse/KEYCLOAK-13370

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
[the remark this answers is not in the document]
slaskawi: In other parts we use arrays (e.g. `UserFederationProviders []KeycloakAPIUserFederationProvider`). I would prefer to keep it like that in all the parts of this PR.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
slaskawi: @vmuzikar The short answer is no. We do a one way sync. See more at https://groups.google.com/g/keycloak-dev/c/6OTpj3ow0u4/m/x2_qCAEEDAAJ

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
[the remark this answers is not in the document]
slaskawi: Similar comment here - let's keep arrays.

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
```diff
@@ -160,6 +160,68 @@ type KeycloakAPIRealm struct {
...
+	Roles *RolesRepresentation `json:"roles,omitempty"`
+}
+
+type RoleRepresentationArray []RoleRepresentation
+
+// https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_rolesrepresentation
+type RolesRepresentation struct {
+	// Client Roles
+	// +optional
+	Client map[string]RoleRepresentationArray `json:"client,omitempty"`
```
slaskawi, line 175: Could we change this to use arrays instead of type redefinition? The Java type is `Map<String, List<RoleRepresentation>>`, so maybe `map[string][]RoleRepresentation` would somehow work? I'm not sure if this makes the Go compiler happy but you probably know what I'm trying to achieve here.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
slaskawi: This sounds good to me.

slaskawi (changes requested):
Ok, it seems we are all on the same page. We just need some updates, a rebase and we're good to go. Please remember to keep the CI green!

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
dhx-teuto: I'd also like that better than what I had to do here.
However `make code/gen` complains
```
.../keycloak-operator/pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go:175:20: map values must be a named type, not *ast.ArrayType
FATA[0000] error generating CRDs from APIs in pkg/apis: error generating CRD manifests: error running CRD generator: not all generators ran successfully 
```
That's the whole reason, also for the discussions above.
The cause is the controller-tools people not wanting this as linked further up https://github.com/kubernetes-sigs/controller-tools/issues/213#issuecomment-507844871
Since we want to replicate the existing data structures as closely as possible this is as close as we can get AFAIK.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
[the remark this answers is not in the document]
dhx-teuto: > the ID in the resource (although respected if present)

correction: while writing the test it was not
This did not cause any new problems, just to clarify in case someone comes across this thread.

dhx-teuto:
rebased, renaming roles is possible, unit tests, e2e tests, linter
From my side this one is good to merge.
@vmuzikar @slaskawi

pkg/apis/keycloak/v1alpha1/keycloakrealm_types.go
slaskawi: Ok, I understand.

slaskawi (approved):
LGTM. Let's give @vmuzikar a day or two to look into this. Then, I'll merge it.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
```diff
@@ -43,9 +43,85 @@ func (i *KeycloakClientReconciler) Reconcile(state *common.ClientState, cr *kc.K
...
+	// note that duplicate role names are impossible thanks to +listType=map
+	for _, role := range rolesMatching {
+		if role.ID == "" {
+			if _, contains := renamedRolesOldNames[role.Name]; contains {
+				desired.AddAction(i.getCreatedClientRoleState(state, cr, role.DeepCopy()))
+			} else {
+				desired.AddAction(i.getUpdatedClientRoleState(state, cr, role.DeepCopy(), role.DeepCopy()))
+			}
+		}
+	}
```
vmuzikar, line 88: I'm not sure I understand why would be ID missing from the CR when doing an update. Or why we would be re-creating the role after rename. The user shouldn't manually remove the ID from the CR. Or maybe I'm just lacking imagination on Monday morning... :D

vmuzikar (commented):
@dhx-teuto Thanks for the update, looks really nice. Added one last comment/question.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
dhx-teuto: Not something I think most people would need, in fact I didn't even consider renaming initially.
But going from name matching to [ID then name] matching comes with a couple of corner cases and instead of failing with weird errors in the operator log I implemented what would be needed to get to the desired state.
Note that the user doesn't necessarily need to "remove" the ID from the CR, if they don't specify it initially we will still run into this because the read state contains IDs (because KC would just generate them) whereas the CR does not (and maybe never did).

vmuzikar: Ok, thanks for clarifying. Good point with the CR missing the ID.

vmuzikar (approved):

dhx-teuto:
Thanks for the discussions! @vmuzikar @slaskawi

d-baranowski:
I added this section to realm crd and nothing happens roles  don't get created :( 

```
    roles:
      realm:
      - name: super user
        id: merlin-super-user
        description: can configure the system, so access to system configuration
        clientRole: false
        composite: false
      - name: merlin-admin
        id: merlin-admin
        description: can create companies, assign account managers to companies
        clientRole: false
        composite: false
      - name: merlin-account manager 
        id: merlin-account-manager
        description: can manage their own accounts
        clientRole: false
        composite: false
```

Kampe:
Yes! Examples would be nice when you all add features.   The examples are severely lacking and I believe are curbing adoption. 

> I added this section to realm crd and nothing happens roles don't get created :(
> 
> ```
>     roles:
>       realm:
>       - name: super user
>         id: merlin-super-user
>         description: can configure the system, so access to system configuration
>         clientRole: false
>         composite: false
>       - name: merlin-admin
>         id: merlin-admin
>         description: can create companies, assign account managers to companies
>         clientRole: false
>         composite: false
>       - name: merlin-account manager 
>         id: merlin-account-manager
>         description: can manage their own accounts
>         clientRole: false
>         composite: false
> ```

PR #296, opened 2020-12-18
Check full commit message for ticket ID

apeschel:
KEYCLOAK-16673

This commit updates the CI for this repository to check the full commit
message for a ticket ID, rather than just the commit title.

### DIFF
--- a/.github/workflows/commit-message.yml
+++ b/.github/workflows/commit-message.yml
@@ -10,4 +10,4 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - name: Check if JIRA ID is in the commit message
-        run: git log -n 1 --pretty=format:'%s' --abbrev-commit | grep -E 'KEYCLOAK-[0-9]+' > /dev/null
+        run: git show -s --format=%B | grep -E 'KEYCLOAK-[0-9]+' > /dev/null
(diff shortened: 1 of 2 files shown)


apeschel:
This is just a quick optional change. Having the ticket ID in titles is going to lead to long commit titles over the standard 50 character limit.

If there's a preference towards keeping the ticket IDs in the title, please close this PR and the corresponding JIRA ticket.

apeschel:
Ticket link, for convenience: https://issues.redhat.com/browse/KEYCLOAK-16673

slaskawi:
@vmuzikar May I ask you to have a look at this one?

@apeschel Could you please create a JIRA for this and remove the OpenAPI correction commit?

vmuzikar:
@slaskawi The JIRA was mentioned here. ;)
> Ticket link, for convenience: https://issues.redhat.com/browse/KEYCLOAK-16673

@apeschel Thank you for your contribution! However, IMHO we don't want to make such change. We should be consistent with other Keycloak repositories that use the same rule – JIRA number as the prefix of commit message.
If we decided to make this change, [CONTRIBUTING.MD](https://github.com/keycloak/keycloak-operator/blob/master/CONTRIBUTING.md) would need to be updated as well and also this change would have to be done across all Keycloak repos.

@slaskawi Could you please confirm this?

slaskawi:
I appreciate your interest in this feature, @apeschel !

I agree with @vmuzikar, we usually use the convention of prefixing the commit title with JIRA ID. You may take a look at other repos for reference. Having said that, let me close this one.

PR #297, opened 2020-12-28
Use a init container to set PVC permissions

Yannig:
Launch a init container to fix permissions on PVC.
Fix a problem when launching a Keycloak instance with aws-ebs disk

## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-16730

## Additional Information

I'm using the Keycloak operator with a Kops cluster deployed in AWS. When I try to deploy a Keycloak instance, the Postgresql instance fail to start with a problem on disk default owner (root).

The only way to fix the problem is to mount this persistent volume, set the owner to postgres (group postgres) and let the postgres container start.

## Verification Steps

- Create a Kops cluster with aws-ebs storage class
- Launch the operator
- Generate a Keycloak instance
- After a few seconds/minutes the Keycloak postgres must be started

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/model/postgresql_deployment.go
+++ b/pkg/model/postgresql_deployment.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


slaskawi:
This probably solves https://issues.redhat.com/browse/KEYCLOAK-14948 too.

slaskawi (changes requested):
Thanks for the contribution, @Yannig !

Could you please change the commit message and get rid of the emoji? Folks are using many different tools for developments and I'm not sure it will render correctly in all of them. Otherwise, LGTM!

Yannig:
Sure, no problem :)

slaskawi (approved):

slaskawi:
Integrated, thanks @Yannig !

PR #298, opened 2021-01-05
KEYCLOAK-16510-Add clientset for Keycloak-Operator

JiyeYu:
Add clientset, lister and informer for Keycloak-Operator.

## JIRA ID
[KEYCLOAK-16510](https://issues.redhat.com/browse/KEYCLOAK-16510)

## Checklist:
- [ ] Automated Tests

## Additional Notes 
This feature has already been discussed with @slaskawi on [Jira](https://issues.redhat.com/browse/KEYCLOAK-16510). 
Most code auto-generated by [k8s.io/code-generator](https://github.com/kubernetes/code-generator).

Example of use:
```
import "github.com/keycloak/keycloak-operator/pkg/client/clientset/versioned"

clientset, err := versioned.NewForConfig(config)
if err != nil {
    log.Fatal(err)
}
keycloak, err = clientset.KeycloakV1alpha1().Keycloaks(namespace).Get(context.TODO(), name, metav1.GetOptions{})
```


slaskawi (approved):
LGTM, may I ask @pb82 @davidffrench @vmuzikar @miquelsi @abstractj to have a look at this also?

slaskawi:
May I ask you to look into this? Are you OK with this changes? @pb82 @davidffrench @vmuzikar @miquelsi @abstractj

vmuzikar:
@slaskawi Will try to look into it today.

davidffrench:
From a quick look @slaskawi it looks good, unfortunately, I won't get a chance to verify the change

vmuzikar (commented):
@JiyeYu Thank you for your contribution!
As far as I can tell looks good to me. Although I haven't deeply checked the auto-generated code.

I've got a couple of questions.
1. Does it make sense to "hide" the generated files – i.e. add them to [`.gitattributes`](https://github.com/keycloak/keycloak-operator/blob/master/.gitattributes)?
2. How to trigger the `code-generator`? Is it part of `make code/gen`?
3. Is this documented somewhere? Or at least is there any explicit mention of the clientset existence in the Operator docs or some in read me? A brief code example (similar to this PR's description) would be enough, IMHO.

JiyeYu:
Hi @vmuzikar 

Thank you for your comment!

> Does it make sense to "hide" the generated files – i.e. add them to .gitattributes?

Thanks, I added it.

> Is this documented somewhere? Or at least is there any explicit mention of the clientset existence in the Operator docs or some in read me? A brief code example (similar to this PR's description) would be enough, IMHO.

There are several documents for code-generator. For example, [its github page](https://github.com/kubernetes/code-generator) provides an [example](https://github.com/kubernetes/sample-controller). You can easily try to use code-generator by [this command](https://github.com/kubernetes/sample-controller/blob/master/hack/update-codegen.sh#L28)

This [article](https://www.openshift.com/blog/kubernetes-deep-dive-code-generation-customresources) describes how to add tags for code-generator.

My experience to use Code-generator is, we'd better to place Keycloak-Operator project in $GOPATH/src. (Then code-generator can easily find the Keycloak API)
Here is my command to call code-generator: 
```
$ cd ${GOPATH}/src/github.com/keycloak/keycloak-operator
$ ${code-generator_dir}/code-generator/generate-groups.sh "client,informer,lister" github.com/keycloak/keycloak-operator/pkg/client github.com/keycloak/keycloak-operator/pkg/apis keycloak:v1alpha1  --go-header-file ./hack/boilerplate.go.txt --output-base "${GOPATH}/src"
```

> How to trigger the code-generator? Is it part of make code/gen?
I tried to place keycloak-operator project outside of $GOPATH/src, some other place outside, and then I tried to use `code-genetator` and nothing happened. (or blank directory with blank files created somewhere.) 
My concern is, code-generator is not easy to use. Integrate it in Makefile may not be a good idea. 
I suggest to update `pkg/client` manually when necessary. (e.g. some big changes happen on apis)

JiyeYu:
Hi @slaskawi 
BTW, I updated [hack/boilerplate.go.txt](https://github.com/keycloak/keycloak-operator/blob/54bc58ed2601fa94362374181129661c7336d955/hack/boilerplate.go.txt), which is used by `code-generator` as a license template.

Basically, I copied the sentences from [prometheus-operator's boilerplate.go.txt](https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/client/versioned/clientset.go#L1).
Both prometheus-operator and keycloak-operator use `Apache License 2.0` License.

If you have any concern on this change, please give me a comment. Thank you.

slaskawi:
@JiyeYu After having another look at this Pull Request, I have a few questions:

- Do we have any community user that wants to use this functionality? I would like to avoid implementing unnecessary things in the Operator as this adds some maintenance burden.
- The only way to make the versioned client maintainable is to (re-)generate it automatically when invoking `make code/gen`. There's no other way.
- It seems the generated client contains fakes. I'm not a big fan of them and I noticed they introduced a few files in our `vendor` directory. Is this necessary? 
- It seems the generated client contains a lot of empty `doc.go` files. Can we remove them?

Finally, can I ask you to start a discussion on this on Keycloak DEV forum? We might get some interesting feedback from our community.

vmuzikar:
@JiyeYu Thank you for your response.

I understand triggering the code-gen is complicated but as @slaskawi already mentioned, we need it to be easy and documented. Also what I am still missing is some mention of the clientset in the docs. Without it users might not be even aware the client exists.

JiyeYu:
@slaskawi @vmuzikar 
Thank you for your comments.

> Do we have any community user that wants to use this functionality? I would like to avoid implementing unnecessary things in the Operator as this adds some maintenance burden.

In fact I want to use this functionality, so I submitted this pull request.
Let me start a discussion on Keycloak DEV.

> The only way to make the versioned client maintainable is to (re-)generate it automatically when invoking make code/gen. There's no other way.

I see. Let me find out a good solution for it. 

> Also what I am still missing is some mention of the clientset in the docs. Without it users might not be even aware the client exists.

For those community users who have the same requirements like me, they can easily check the source code to find out if Keycloak Operator provides `clientsets`. Of course a doc is better for users to know this feature.

> It seems the generated client contains fakes. I'm not a big fan of them and I noticed they introduced a few files in our vendor directory. Is this necessary?
It seems the generated client contains a lot of empty doc.go files. Can we remove them?

Fakes, empty doc.go and new files in vendor directory are all automatically generated by code-generator. So if we want to remove them, we need to write a script to do this job after code/gen every time.

slaskawi:
@JiyeYu Could I ask you to address the issues me mentioned in Keycloak DEV? I'm especially interested in generating the Clientset with `make code/gen`.

JiyeYu:
Sorry I I didn't try it these days. I will implement it soon

JiyeYu:
Hi @slaskawi @vmuzikar,

Sorry to keep you waiting. 

I figure out how to use code-generator under mode GO111MODULE="on"
 (that means git clone Keycloak-Operator in some other directory, not in GOPATH. For example, I usually clone Keycloak-Operator into ~/workspace/kubernetes/keycloak/keycloak-operator; But my GOPATH is ~/go)

As a matter of fact, project code-generator is hard to use. It is not easy to implement the automation by `make client/gen`.

**Before I implement the patch, please let me make a confirmation. Sorry for the long comment.**

### 1. Is it acceptable to add k8s.io/code-generator into Keycloak-Operator/vendor

Your advice is to install a bin of code-generator by `go get `
However, according to [this](https://github.com/kubernetes/code-generator/issues/92#issuecomment-610795582), code-generator cannot be built. 

Check the project, we can find the usage of code-generator is: 
```
code-generator/generate-groups.sh <generators> <output-package> <apis-package> <groups-versions> ...

  <generators>        the generators comma separated to run (deepcopy,defaulter,client,lister,informer) or "all".
  <output-package>    the output package name (e.g. github.com/example/project/pkg/generated).
  <apis-package>      the external types dir (e.g. github.com/example/api or github.com/example/project/pkg/apis).
  <groups-versions>   the groups and their versions in the format "groupA:v1,v2 groupB:v1 groupC:v2", relative
                      to <api-package>.
```

According to by code-generator's example, [sample-controller](https://github.com/kubernetes/sample-controller#when-using-go-111-modules), integrate code-generator into vendor folder is a highly recommended.

### 2. Is it acceptable that the script needs to execute `cp -r <output client not in correct directory> <keycloak-operator/pkg/client> after clientset generated.

Take sample-controller as an example, example command will be:

```
generate-groups.sh 
  "deepcopy,client,informer,lister" \                  # <generators>
  k8s.io/sample-controller/pkg/client \             # <output-package>
  k8s.io/sample-controller/pkg/apis \               # <apis-package>
  samplecontroller:v1alpha1 \                            # <groups-versions>
  --output-base "$../../.."
```

Then, the output will be output in directory:
`~/workspace/k8s.io/sample-controller/pkg/client`

while sample-controller is cloned in directory: 
`~/workspace/kubernetes/sample-controller`

What I mean is if the parent directory of sample-controller is not `k8s.io`,  new directory `k8s.io/sample-controller/pkg/client` will be created. Then we need to cp the client directory to `sample-controller` project.

 We even cannot change <output-package> to `sample-controller/pkg/client`. 
If we do this way, the import part of output file will be incorrect:

```
--- a/pkg/generated/clientset/versioned/clientset.go
+++ b/pkg/generated/clientset/versioned/clientset.go
@@ -20,11 +20,11 @@ package versioned

 import (
        "fmt"
+       samplecontrollerv1alpha1 "sample-controller/pkg/generated/clientset/versioned/typed/samplecontroller/v1alpha1"
-       samplecontrollerv1alpha1 "k8s.io/sample-controller/pkg/generated/clientset/versioned/typed/samplecontroller/v1alpha1"
 )
```

### Summary
code-generator is useful, when we insist to use it by script under mode GO111MODULE="on", 
it is not decent.

slaskawi:
> I figure out how to use code-generator under mode GO111MODULE="on"

@JiyeYu Yes, we develop the Operator with an assumption this options is on. If I remember correctly, we used to set it in our Makefile for Go 12. Then we migrated to 13 and this option was removed from our Makefile. 

> 1. Is it acceptable to add k8s.io/code-generator into Keycloak-Operator/vendor?

I'm sorry to bring you bad news, but this is a no-go for me. Once a dependency gets into our `vendor` it adds a lot of maintenance burden. In most of the time, I'm trying to do the opposite thing - trim as many dependencies as I can to make the project slimmer.

Isn't there any other way to do it?

> 2. Is it acceptable that the script needs to execute cp -r <output-directory>/client keycloak-operator/pkg/client after clientset generated?

As long as it automated in `make code/gen`, I see no problem.

JiyeYu:
@slaskawi 
I see. No problem. There is some other way to use code-generator.
My idea is using git clone to download code-generator in folder ./tmp
After generating clients, folder code-generator will be deleted.

Makefile
```diff
@@ -179,6 +179,18 @@ code/lint:
 	@echo "--> Running golangci-lint"
 	@$(shell go env GOPATH)/bin/golangci-lint run --timeout 10m
 
+.PHONY: client/gen
+client/gen:
+	@echo "--> Running code-generator to generate clients"
```
slaskawi, line 184: @pskopek @ASzc Could you please have a look at this? Is this acceptable for our pipelines?

slaskawi (dismissed):
LGTM but I need an additional confirmation from @ASzc and @pskopek

slaskawi:
@ASzc @pskopek @miquelsi @abstractj @vmuzikar Last chance gentlemen? I plan to merge this soon.

Makefile
```diff
@@ -179,6 +179,18 @@ code/lint:
...
+	# prepare tool code-generator
+	@mkdir -p ./tmp/code-generator
+	@git clone https://github.com/kubernetes/code-generator.git ./tmp/code-generator
```
vmuzikar, line 187: Should we point to a specific commit/tag rather than rely on `master`?

Makefile
```diff
@@ -179,6 +179,18 @@ code/lint:
 	@echo "--> Running golangci-lint"
 	@$(shell go env GOPATH)/bin/golangci-lint run --timeout 10m
 
+.PHONY: client/gen
+client/gen:
```
vmuzikar, line 183: We should document this in [README.md](https://github.com/keycloak/keycloak-operator/blob/2d19aae251da8f815fa00b370ea5c027cd994324/README.md#L118).

Makefile
vmuzikar: I think it won't be a problem since it is not run by default (correct?) and therefore won't be ever executed in the pipelines.

vmuzikar (changes requested):
@JiyeYu Thanks for the updates! Added a few comments.

Makefile
vmuzikar: But this brings me to another question. Should we hook up this target with `code/gen`? I'm a bit afraid it'll get forgotten and we end up with outdated client...

JiyeYu:
@slaskawi @vmuzikar 

Thank you for your message. I have made some changes according to them.

> But this brings me to another question. Should we hook up this target with code/gen? I'm a bit afraid it'll get forgotten and we end up with outdated client...

I haven't made the hook. Do I need to do it? Waiting for your decision.

slaskawi:
> I haven't made the hook. Do I need to do it? Waiting for your decision.

@JiyeYu Yes please. The easiest way to do it is to use `code/gen: client/gen`.

After implementing this, this PR LGTM. @vmuzikar any last comments?

JiyeYu:
@slaskawi  @vmuzikar 

Thanks for the comments. I have made the hook. Now code/gen will do client/gen at the same time.

vmuzikar (approved):
@JiyeYu Thank you for the update! LGTM.

slaskawi (approved):

slaskawi:
Integrated, thanks @JiyeYu !

Makefile
ASzc: Gmail relegated this to the Forums tab for me (a tab with very high noise ratio), so I never saw it. Doing code generation in the pipelines is a big problem, as we've seen with operator-sdk. It really should just be a rule that this is run by the developers for each change, before handing off the code. If the code isn't right, that's not the build pipeline's problem.

PR #299, opened 2021-01-06
fix(alerts): remove the KeycloakJavaNonHeapThresholdExceeded alert

wei-lee:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16751

## Additional Information

There is no action that SRE can take when this alert is firing, and in prod environment it has been fired a lot and generated a lot of false positives.

Another option could be we can increase the time for which the alert will be generated. But again, when the alert is fired, there is no action that SRE can take to solve the issue.

## Checklist:
- [ ] Automated Tests


wei-lee:
@slaskawi can you review this please?

slaskawi (approved):

slaskawi:
Integrated, thanks @wei-lee !

PR #301, opened 2021-01-07
Check route readiness only if ExternalAccess is enabled

matskiv:
## JIRA
https://issues.redhat.com/browse/KEYCLOAK-16771

## Additional Information
*What:*
On OpenShift, Keycloak CR will be stuck with `status.ready = false` if `.spec.ExternalAccess.enable = false`

*Why:*
Keycloak Operator should not be waiting for the route to become ready if the route is disabled via Keycloak CR spec.

## Verification Steps
1. Install the operator on a testing Openshift cluster (or run it locally against a namespace on Openshift cluster)
2. Create a valid Keycloak CR with `.spec.ExternalAccess.enable = false`
3. After keycloak StatefulSet and postgres deployment become ready, `.status.ready` field of the CR should become `true` 

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/common/cluster_state.go
+++ b/pkg/common/cluster_state.go
@@ -483,10 +483,12 @@ func (i *ClusterState) IsResourcesReady(cr *kc.Keycloak) (bool, error) {
 	}
 
 	// If running on OpenShift, check the Route is ready
-	stateManager := GetStateManager()
-	openshift, keyExists := stateManager.GetState(RouteKind).(bool)
-	if keyExists && openshift {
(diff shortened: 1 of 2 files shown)


slaskawi (changes requested):
Could you add a test for this case? It might be a unit test. I just want to make sure we won't break it again.

matskiv:
@slaskawi good idea. Test added.

slaskawi (approved):

slaskawi:
Integrated, thanks @matskiv !

PR #307, opened 2021-02-08
KEYCLOAK-15254 Use master or latest images for development

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-15254

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "9.0.2"
+	Version = "master"
 )
--- a/pkg/model/image_manager.go
+++ b/pkg/model/image_manager.go
@@ -15,7 +15,7 @@ const (
 	RHMIBackupContainer   = "RELATED_IMAGE_RHMI_BACKUP_CONTAINER"
 	PostgresqlImage       = "RELATED_IMAGE_POSTGRESQL"
 
-	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:9.0.2"
+	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:latest"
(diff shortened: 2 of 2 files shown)


version/version.go
```diff
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "9.0.2"
+	Version = "master"
```
vmuzikar, line 4: If I understand it correctly, this represents the operator version. Not sure if we shouldn't use something like `$NEXT_VERSION-SNAPSHOT` like we do for other Keycloak projects.

miquelsi (approved):

version/version.go
slaskawi: At the moment, we push the latest bits to "master" tag in Quay. If we changed the version here, we should probably be consistent. Then, using "13.0.0-SNAPSHOT" tag in container world looks awkward. 

Maybe this is something we should get back when constructing end-to-end CI/CD pipeline? Right now using "master" seems the most natural thing, as the image tag is "master" and we're consuming Keycloak image with the "master" tag.

vmuzikar: It raised a flag for me because even non-Java KC projects use  a versioning in `master`, e.g. the Node.js adapter [uses `-dev` sufix](https://github.com/keycloak/keycloak-nodejs-connect/blob/f8e011aea5021b81cd085df078eee7b5d773d0cc/package.json#L3). And KC server images also [use a similar `latest` tag](https://quay.io/repository/keycloak/keycloak?tab=tags) which does not represent a specific version. But I don't know the Go/images best practices, so let's leave it as is.

vmuzikar (approved):

abstractj (approved):

PR #310, opened 2021-02-15
KEYCLOAK-16548 Collect more Keycloak metrics

slaskawi:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-16548

## Additional Information

This Pull Request creates additional Service for gathering Wildfly metrics. However, the most important change is to the Dashboard. Here's how it looks now [link](https://snapshot.raintank.io/dashboard/snapshot/fcIxz4Iu6qwuVhI3EU3JcgUtS2bWmvrU). Note, that some graphs didn't render correctly because RHSSO doesn't provide any way to enable metrics (we probably need to implement it - checking with @iankko).

image

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/common/auto_detect.go
+++ b/pkg/common/auto_detect.go
@@ -1,8 +1,6 @@
 package common
 
 import (
-	"k8s.io/apimachinery/pkg/runtime/schema"
-
 	"time"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
@@ -15,9 +13,8 @@ import (
 
 // Background represents a procedure that runs in the background, periodically auto-detecting features
 type Background struct {
-	dc                  discovery.DiscoveryInterface
-	ticker              *time.Ticker
-	SubscriptionChannel chan schema.GroupVersionKind
+	dc     discovery.DiscoveryInterface
+	ticker *time.Ticker
 }
 
 // New creates a new auto-detect runner
@@ -27,20 +24,16 @@ func NewAutoDetect(mgr manager.Manager) (*Background, error) {
 		return nil, err
 	}
 
-	// Create a new channel that GVK type will be sent down
-	subChan := make(chan schema.GroupVersionKind, 1)
-
-	return &Background{dc: dc, SubscriptionChannel: subChan}, nil
+	return &Background{dc: dc}, nil
 }
 
 // Start initializes the auto-detection process that runs in the background
 func (b *Background) Start() {
+	b.autoDetectCapabilities()
 	// periodically attempts to auto detect all the capabilities for this operator
 	b.ticker = time.NewTicker(5 * time.Second)
 
 	go func() {
-		b.autoDetectCapabilities()
-
 		for range b.ticker.C {
 			b.autoDetectCapabilities()
 		}
@@ -50,7 +43,6 @@ func (b *Background) Start() {
 // Stop causes the background process to stop auto detecting capabilities
 func (b *Background) Stop() {
 	b.ticker.Stop()
-	close(b.SubscriptionChannel)
 }
 
 func (b *Background) autoDetectCapabilities() {
@@ -65,29 +57,22 @@ func (b *Background) detectRoute() {
 		// Set state that the Route kind exists. Used to determine when a route or an Ingress should be created
 		stateManager := GetStateManager()
 		stateManager.SetState(RouteKind, true)
-
-		b.SubscriptionChannel <- routev1.SchemeGroupVersion.WithKind(RouteKind)
 	}
 }
 
 func (b *Background) detectMonitoringResources() {
 	// detect the PrometheusRule resource type exist on the cluster
+	stateManager := GetStateManager()
 	resourceExists, _ := k8sutil.ResourceExists(b.dc, monitoringv1.SchemeGroupVersion.String(), monitoringv1.PrometheusRuleKind)
-	if resourceExists {
-		b.SubscriptionChannel <- monitoringv1.SchemeGroupVersion.WithKind(monitoringv1.PrometheusRuleKind)
-	}
+	stateManager.SetState(monitoringv1.PrometheusRuleKind, resourceExists)
 
 	// detect the ServiceMonitor resource type exist on the cluster
 	resourceExists, _ = k8sutil.ResourceExists(b.dc, monitoringv1.SchemeGroupVersion.String(), monitoringv1.ServiceMonitorsKind)
-	if resourceExists {
-		b.SubscriptionChannel <- monitoringv1.SchemeGroupVersion.WithKind(monitoringv1.ServiceMonitorsKind)
-	}
+	stateManager.SetState(monitoringv1.ServiceMonitorsKind, resourceExists)
 
 	// detect the GrafanaDashboard resource type resourceExists on the cluster
 	resourceExists, _ = k8sutil.ResourceExists(b.dc, grafanav1alpha1.SchemeGroupVersion.String(), grafanav1alpha1.GrafanaDashboardKind)
-	if resourceExists {
-		b.SubscriptionChannel <- grafanav1alpha1.SchemeGroupVersion.WithKind(grafanav1alpha1.GrafanaDashboardKind)
-	}
+	stateManager.SetState(monitoringv1.ServiceMonitorsKind, resourceExists)
 }
 
 func (b *Background) detectOpenshift() {
--- a/pkg/controller/keycloak/keycloak_reconciler_test.go
+++ b/pkg/controller/keycloak/keycloak_reconciler_test.go
@@ -54,11 +54,12 @@ func TestKeycloakReconciler_Test_Creating_All(t *testing.T) {
 	//    7) Postgresql Service
 	//    8) Keycloak Service
 	//    9) Keycloak Discovery Service
-	//    10) Keycloak Probe ConfigMap
-	//    11) Keycloak StatefulSets
-	//    12) Keycloak Route
+	//    10) Keycloak Monitoring Service
+	//    11) Keycloak Probe ConfigMap
+	//    12) Keycloak StatefulSets
 	//        Note, that's no MetricsRoute as it needs an established hostname from the root route
-	assert.Equal(t, len(desiredState), 13)
+	//    13) Keycloak Route
+	assert.Equal(t, len(desiredState), 14)
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[0])
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[1])
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[2])
@@ -72,6 +73,7 @@ func TestKeycloakReconciler_Test_Creating_All(t *testing.T) {
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[10])
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[11])
 	assert.IsType(t, common.GenericCreateAction{}, desiredState[12])
+	assert.IsType(t, common.GenericCreateAction{}, desiredState[13])
 	assert.IsType(t, model.KeycloakAdminSecret(cr), desiredState[0].(common.GenericCreateAction).Ref)
 	assert.IsType(t, model.PrometheusRule(cr), desiredState[1].(common.GenericCreateAction).Ref)
 	assert.IsType(t, model.ServiceMonitor(cr), desiredState[2].(common.GenericCreateAction).Ref)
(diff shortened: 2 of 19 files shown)


iankko:
@slaskawi 

> Note, that some graphs didn't render correctly because RHSSO doesn't provide any way to enable metrics (we probably need to implement it - checking with @iankko).

See the email reply (IMHO it's just question of choosing an option, which best suits your purpose):
> If there's no dedicated option - is there an option to break into the container in a nice way? :) I can think of some dirty tricks as replacing the bootstrap script etc. But that's not nice, and complicated from the upgrade perspective...
```
Looked, but there doesn't seem to be a dedicated option available so far, AFAICT. They could be enabled via three ways AFAICT.
Listing them in order of difficulty / time demands necessary to implement them:

1) Using CLI extensions - if it's enough just to switch the "statistics-enabled" property on some datasource from "false" to "true" and this is doable via JBoss CLI, then it should be feasible to achieve via extensions.cli (but didn't try):

https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/modules/sso/sso-extensions-cli/added/configure_sso_cli_extensions.sh#L7

2) Second way is to define custom definition of the datasource in question (but more complex IMHO than 1) ):

https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.3/html-single/getting_started_with_jboss_eap_for_openshift_online/index#building-custom-layers_default

(scroll down to "postgresql-datasource" definition)

3) If none of the above would work, we could add a dedicated option to have it enabled. Seems we are still using a dedicated
(overridden datasource code), derived from EAP:

https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/modules/sso/config/launch/setup/74/added/launch/datasource.sh#L8

Thus could add the new option / variable there. But this would take the longest time to achieve (implement + PR review
+ time till this gets into the next RH-SSO release).

Hope (at least some of the above) helps
Jan
```

slaskawi:
Thanks for the explanation @iankko. We probably need to go with option number 3.

pedroigor:
@slaskawi I remember that one issue we had was related to the `Number of entries in caches` not showing the information. I remember we discussing about issues in ISPN subsystem statistics/metrics ... Not sure.

/cc @sguilhen

david-martin:
👍  @slaskawi This looks great.
1 question about the HTTP Errors panel.
How does it look when there are a lot of different routes and codes to sum by?
Will the route contain any uuids?

Rajagopalan-Ranganathan:
@slaskawi : Can the panels be rearranged so that we don't have many empty spaces. Also, should try to rearrange them with the thought of which of those metrics/panels are of great importance to SRE. (Cosmetic change, but the goal is minimize scrolling and SRE people getting the important information at the top of their screen).

Rajagopalan-Ranganathan:
@slaskawi a Follow up issue/task could be to create more specific dashboards that can be landed from a drill-down link in the main dashboard (this would keep the top level dashboard simple and SRE can then visit the specific dashboard for more details). Ex: Aggregated use of compute metrics shown in landing dashboard and drill-down link takes to specific pods/namespaces/ e.t.c

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -54,11 +54,11 @@ func TestKeycloakReconciler_Test_Creating_All(t *testing.T) {
 	//    7) Postgresql Service
 	//    8) Keycloak Service
 	//    9) Keycloak Discovery Service
-	//    10) Keycloak Probe ConfigMap
-	//    11) Keycloak StatefulSets
-	//    12) Keycloak Route
+	//    10) Keycloak Monitoring Service
+	//    11) Keycloak Probe ConfigMap
+	//    12) Keycloak StatefulSets
 	//        Note, that's no MetricsRoute as it needs an established hostname from the root route
```
Rajagopalan-Ranganathan, line 60: Comment the 13th item here.

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -54,11 +54,11 @@ func TestKeycloakReconciler_Test_Creating_All(t *testing.T) {
 	//    7) Postgresql Service
 	//    8) Keycloak Service
 	//    9) Keycloak Discovery Service
-	//    10) Keycloak Probe ConfigMap
-	//    11) Keycloak StatefulSets
-	//    12) Keycloak Route
+	//    10) Keycloak Monitoring Service
+	//    11) Keycloak Probe ConfigMap
+	//    12) Keycloak StatefulSets
```
Rajagopalan-Ranganathan, line 59: Can we add the comment for the newly added 13th (14th ) item here?

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -54,11 +54,11 @@ func TestKeycloakReconciler_Test_Creating_All(t *testing.T) {
...
 	//        Note, that's no MetricsRoute as it needs an established hostname from the root route
-	assert.Equal(t, len(desiredState), 13)
+	assert.Equal(t, len(desiredState), 14)
```
Rajagopalan-Ranganathan, line 61: Would be nice to have the length of the dseried state defined as a constant. Easier to read, change and maintainable. Single line change in the future.

slaskawi:
> Can the panels be rearranged so that we don't have many empty spaces. Also, should try to rearrange them with the thought of which of those metrics/panels are of great importance to SRE. (Cosmetic change, but the goal is minimize scrolling and SRE people getting the important information at the top of their screen).

@Rajagopalan-Ranganathan The panels are design to reside in one of the 3 sections: System health, Detailed Metrics and Experimental. The key design principles are exactly those as you mention - the System health panel enables the SRE Team to quickly tell if the service is healthy or not. The detailed metrics enables further diagnosis. The dashboard fits into a larger [Observability Proposal](https://github.com/keycloak/keycloak-community/pull/177).

So to sum it up - being an SRE Team you should probably never scroll the dashboard down. Being a developer, who investigates the root cause - you will probably need to go though most of the panels.

> a Follow up issue/task could be to create more specific dashboards that can be landed from a drill-down link in the main dashboard (this would keep the top level dashboard simple and SRE can then visit the specific dashboard for more details). Ex: Aggregated use of compute metrics shown in landing dashboard and drill-down link takes to specific pods/namespaces/ e.t.c

@Rajagopalan-Ranganathan So what exactly is missing? Perhaps this is something we could discuss first in the Observability proposal? https://github.com/keycloak/keycloak-community/pull/177

Rajagopalan-Ranganathan:
> > Can the panels be rearranged so that we don't have many empty spaces. Also, should try to rearrange them with the thought of which of those metrics/panels are of great importance to SRE. (Cosmetic change, but the goal is minimize scrolling and SRE people getting the important information at the top of their screen).
> 
> @Rajagopalan-Ranganathan The panels are design to reside in one of the 3 sections: System health, Detailed Metrics and Experimental. The key design principles are exactly those as you mention - the System health panel enables the SRE Team to quickly tell if the service is healthy or not. The detailed metrics enables further diagnosis. The dashboard fits into a larger [Observability Proposal](https://github.com/keycloak/keycloak-community/pull/177).
> 
> So to sum it up - being an SRE Team you should probably never scroll the dashboard down. Being a developer, who investigates the root cause - you will probably need to go though most of the panels.
> 
> > a Follow up issue/task could be to create more specific dashboards that can be landed from a drill-down link in the main dashboard (this would keep the top level dashboard simple and SRE can then visit the specific dashboard for more details). Ex: Aggregated use of compute metrics shown in landing dashboard and drill-down link takes to specific pods/namespaces/ e.t.c
> 
> @Rajagopalan-Ranganathan So what exactly is missing? Perhaps this is something we could discuss first in the Observability proposal? [keycloak/keycloak-community#177](https://github.com/keycloak/keycloak-community/pull/177)
@slaskawi My thought was to may be move some of the panels ( like the ones the developer is interested and not the SRE) and have them as separate dashboards so that a deep link from the landing/main dashboard would take us to more detailed metrics dashboards. this would keep the main dashboard simple for the SRE.

slaskawi:
> My thought was to may be move some of the panels ( like the ones the developer is interested and not the SRE) and have them as separate dashboards so that a deep link from the landing/main dashboard would take us to more detailed metrics dashboards. this would keep the main dashboard simple for the SRE.

@Rajagopalan-Ranganathan ooohhh no - this might not be a good idea! Please remember that one of the most powerful features is to limit the time range (or "zoom the dashboard in") and deeply investigate what happened in a specific period of time. Then you definitely need information about readiness probes, latency etc on a single dashboard and you definitely don't want to jump between the multiple ones.

Just to give you a specific example, if your memory gets full, you will first notice an increased overall latency in the System health panel (the top one). Then you can zoom the dashboard in and move down noticing higher CPU usage and higher GC rate. Those are intentionally next to each other. Finally, you move your eyes left to the memory usage and you have full information about what's going on.

Maybe we could tweak placement for some of those graphs but it requires some thought. In general the approach is: go top to bottom and left to right to gather details on what's going on.

pkg/controller/keycloak/keycloak_reconciler_test.go
slaskawi: Done.

pkg/controller/keycloak/keycloak_reconciler_test.go
slaskawi: This is one of the problems that we have with no good solution to be honest. 

On one hand, we decided to test the reconciler using unit tests. That decision turned out to work really great. On the other hand, we need additional "safety net" to verify if the reconciler creates all the objects we need in proper order. An alternative would be to use integration tests for, which is even worse.

So to get back to your suggestion - I don't think using a variable in this case makes sense. The variable would be used in only one place - to verify if the number of emitted objects is fine.

Rajagopalan-Ranganathan: @slaskawi After a quick look, a top of my mind suggestion would be, create a method to generate/give a `common.DesiredClusterState` that would be expected for the test case. And then compare the expected object (returned by the method) to the one returned by the reconciler. This should eliminate the need for index by index assertion of expected things. Would this be OK? Am i missing something?

slaskawi: That's a neat idea. Let me create a JIRA for this.

Rajagopalan-Ranganathan:
> > My thought was to may be move some of the panels ( like the ones the developer is interested and not the SRE) and have them as separate dashboards so that a deep link from the landing/main dashboard would take us to more detailed metrics dashboards. this would keep the main dashboard simple for the SRE.
> 
> @Rajagopalan-Ranganathan ooohhh no - this might not be a good idea! Please remember that one of the most powerful features is to limit the time range (or "zoom the dashboard in") and deeply investigate what happened in a specific period of time. Then you definitely need information about readiness probes, latency etc on a single dashboard and you definitely don't want to jump between the multiple ones.
> 
> Just to give you a specific example, if your memory gets full, you will first notice an increased overall latency in the System health panel (the top one). Then you can zoom the dashboard in and move down noticing higher CPU usage and higher GC rate. Those are intentionally next to each other. Finally, you move your eyes left to the memory usage and you have full information about what's going on.
> 
> Maybe we could tweak placement for some of those graphs but it requires some thought. In general the approach is: go top to bottom and left to right to gather details on what's going on.

Thanks @slaskawi it makes sense now. We can rearrange the panels after receiving feedback from the AppSRE people as well. I am sure they have some thoughts on this.

pkg/controller/keycloak/keycloak_reconciler_test.go
Rajagopalan-Ranganathan: @slaskawi thanks a lot!

pkg/common/auto_detect.go
```diff
@@ -1,8 +1,6 @@
 package common
 
 import (
-	"k8s.io/apimachinery/pkg/runtime/schema"
```
vmuzikar, line 4: Just out of curiosity, why are we removing this package across the project?

pkg/model/keycloak_deployment.go
```diff
@@ -152,6 +152,10 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
 			Name:  "PROXY_ADDRESS_FORWARDING",
 			Value: "true",
 		},
+		{
+			Name:  "KEYCLOAK_STATISTICS",
```
vmuzikar, line 156: Is this variable documented somewhere? I suppose it's a var used by KC, not by the operator.

vmuzikar (approved):
As far as I can tell (this is a bit out of scope of my expertise :)), looks good to me but I didn't try it. Added two questions.

pkg/common/auto_detect.go
slaskawi: I simplified the auto detecting mechanism. It turned out it contained a race condition (between the reconciler and the channel in the `auto_detect.go`). Once I did that, the schema started to be useless. 

I plan to simplify it a bit more in the future. I'd like to squash Auto Detect mechanism with the Background object we have. Currently, it's hard to wrap your head around it...

pkg/model/keycloak_deployment.go
slaskawi: Here's the manual for this: https://github.com/keycloak/keycloak-containers/tree/master/server#enable-some-metrics

In the Operator, we need the metrics for Grafana.

pedroigor (dismissed):

pedroigor (approved):

PR #311, opened 2021-02-15
Add Pod Antiaffinity to keycloak deployment

Rajagopalan-Ranganathan:
PodAntiaffinity settings are added to keycloak deployment spec. When the
affinity values are provided in the CR it is used, else the default values are used.

## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16482
https://issues.redhat.com/browse/MGDSTRM-1658

## Additional Information

Default podantiaffinity settings that would be used if none specified:
``` affinity:
    podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: app
              operator: In
              values:
              -keycloak
          topologyKey: "topology.kubernetes.io/zone"
      - weight: 90
        podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: app
              operator: In
              values:
              - keycloak
          topologyKey: "kubernetes.io/hostname"
```

1. Use the example as mentioned above or create a new spec with podantiaffinity specified
2. Deploy the keycloak CR (from the above step)
3. Verify that the keycloak pod is created and running successfully
4. Create a multi AZ cluster, deploy the keycloak and see if it works

## Checklist:
- [x] Automated Tests


Rajagopalan-Ranganathan:
@slaskawi  - Please take a look

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -79,6 +79,9 @@ type KeycloakSpec struct {
 	// Name of the StorageClass for Postgresql Persistent Volume Claim
 	// +optional
 	StorageClassName *string `json:"storageClassName,omitempty"`
+	//POD affinity rules for KeyCloak deployment
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
```
slaskawi, line 84: I think we should probably hide the complexity under some nice option. Perhaps something like this:
```
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  externalAccess:
    enabled: True
  nodeAntiAffinity:
    enabled: True
```

Thinking about this more, I think both `nodeAntiAffinity` and `podDisruptionBudget` should be always set to true (WDYT @wei-lee). It's hard to think about the scenarios where you don't want these.

slaskawi (changes requested):

pkg/apis/keycloak/v1alpha1/keycloak_types.go
Rajagopalan-Ranganathan: @slaskawi would this then reduce the flexibility for the user to specify his own settings/config/preferences for these? (in case he needs them).  Surely our default config cannot satisfy everyone's needs.

slaskawi: I don't think it matters in this case. Multi AZ deployment are all similar.

But if you insist, we can make a compromise here. Let's introduce a `nodeAntiAffinity` setting as well as raw Affinity settings in the experimental section. Would that work for you?

Rajagopalan-Ranganathan: @slaskawi are we always mandating Multi AZ deployment? (sorry for my ignorance).
Curious to understand why we want them in the experimental section ( i assumed experimental section can change/be removed in the future). Wouldn't these affinity settings be more of a standard requirement if desired?

Rajagopalan-Ranganathan: After discussing with @slaskawi - Decided to hide the low level details from the user. The deployment CR would have ```  multiAvailablityZones:
    enabled: True``` that can enabled for multi Active zone deployments. This would create a keycloak deployment with the podAntiAffinity settings.

pkg/model/keycloak_deployment.go
```diff
@@ -417,3 +422,46 @@ func readinessProbe() *v1.Probe {
 		FailureThreshold:    ProbeFailureThreshold,
 	}
 }
+
+func KeycloakPodAffinity(cr *v1alpha1.Keycloak) *v1.Affinity {
+	return &v1.Affinity{
+		PodAffinity: &v1.PodAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{
```
slaskawi, line 429: @ryanemerson Could I ask you to have a look at this? LGTM but having a double check won't hurt ;)

slaskawi (approved):
LGTM! Great work @Rajagopalan-Ranganathan !

Let's give Ryan a day or two for the affinity settings review. Once this happens, I'll merge this one.

pkg/model/keycloak_deployment.go
```diff
@@ -417,3 +422,46 @@ func readinessProbe() *v1.Probe {
 		FailureThreshold:    ProbeFailureThreshold,
 	}
 }
+
+func KeycloakPodAffinity(cr *v1alpha1.Keycloak) *v1.Affinity {
+	return &v1.Affinity{
+		PodAffinity: &v1.PodAffinity{
```
ryanemerson, line 428: Don't you want `PodAntiAffinity` here to ensure that keycloak deploys to different zones and nodes?

e.g.:

```go
&v1.Affinity{
  PodAntiAffinity: &v1.PodAntiAffinity{
    PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{{
```

pkg/apis/keycloak/v1alpha1/keycloak_types.go
```diff
@@ -79,6 +79,9 @@ type KeycloakSpec struct {
 	// Name of the StorageClass for Postgresql Persistent Volume Claim
 	// +optional
 	StorageClassName *string `json:"storageClassName,omitempty"`
+	// Specify PodAntiAffinity settings for Keycloak deployment in Multi AZ
+	// +optional
+	MultiAvailablityZones MultiAvailablityZonesConfig `json:"multiAvailablityZones,omitempty"`
```
ryanemerson, line 84: Something to consider. The approach we took in Infinispan was to allow users to define their own policies by exposing the Affinity spec on our CR, e.g.:

```go
type InfinispanSpec struct {
	...
	Affinity  *corev1.Affinity        `json:"affinity,optional,omitempty"`
}
```

This allows users to define their own Affinity or AntiAffinity policies if they have additional requirements (co locating keycloak/infinispan pods on the same nodes as a performance intensive application for example). In our case, if the user doesn't specify a policy then we provide a default strategy. In our documentation we then have examples of different use-cases and appropriate Affinity strategies.

Here's the Infinispan PR if you're curious https://github.com/infinispan/infinispan-operator/pull/654

pkg/model/keycloak_deployment.go
Rajagopalan-Ranganathan: Good catch, It should be `podAntiAffinity` , I will update the PR now. (IDE autocompletion was overlooked).

pkg/apis/keycloak/v1alpha1/keycloak_types.go
slaskawi: Didn't we agree to put raw Affinity object in the experimental section?

pkg/apis/keycloak/v1alpha1/keycloak_types.go
Rajagopalan-Ranganathan: Myself and @slaskawi had a discussion around this, the first version of PR had the changes this way. Then @slaskawi mentioned/suggested that, we wanted to avoid users configuring low level knobs of the system and wanted to keep the top level keycloak CR as simple as possible. That is the reason we removed the affinity settings from the CR, and kept a boolean enablement feature.

pkg/model/keycloak_deployment.go
Rajagopalan-Ranganathan: Fixed.

pkg/apis/keycloak/v1alpha1/keycloak_types.go
Rajagopalan-Ranganathan: @slaskawi yes, i missed that. I will update the PR now.

Rajagopalan-Ranganathan: @slaskawi  : in a scenario where we have `multiAvailablityZones: enabled: True` and the user had defined the affinity object in the Experimental section, which one should be used (takes precedence?), I think it should be the explicit settings done by the user. Is that correct?

ryanemerson (approved):

pkg/apis/keycloak/v1alpha1/keycloak_types.go
Rajagopalan-Ranganathan: Agreed with @slaskawi that the explicit settings done for affinity in the experimental section would take precedence even in the case of multiAvalabiltyZones being enabled.

Rajagopalan-Ranganathan:
@ryanemerson @slaskawi - Made the changes for adding affinity object in experimental section. Added unit tests as well. Please review when you find time.

ryanemerson:
LGTM

pkg/apis/keycloak/v1alpha1/keycloak_types.go
slaskawi: Works for me! Thanks!

slaskawi (approved):

slaskawi:
LGTM, would you mind to rebase this @Rajagopalan-Ranganathan ? Once this is done, I'll merge it.

slaskawi (approved):

PR #314, opened 2021-02-19
KEYCLOAK-16003 Add support for mounting secrets

soleblaze:
This adds experimental support to mount secrets as volumes the same way
that configmaps can be mounted.

## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16003

## Additional Information
This is needed to be able to mount keytabs for kerberos support. Since keytabs are considered sensitive files they should be stored in a secret instead of a configmap. 

## Verification Steps

1. Create a secret named keytab that contains a test key:value
2. Create a keycloak instance using the following, or something similiar (as seen in deploy/examples/keycloak/keycloak-with-experimental-settings.yaml):
  
```
kind: Keycloak
metadata:
  name: keycloak
  labels:
    app: sso
spec:
  instances: 1
  externalAccess:
    enabled: True
  keycloakDeploymentSpec:
    experimental:
      volumes:
        defaultMode: 0777
        items:
          - secret:
              name: keytabs
              mountPath: /keytabs
```

3. Validate that keycloak-0 came up properly and mounted the keytabs secret onto the /keytabs path.

## Checklist:
- [ ] Automated Tests


deploy/examples/keycloak/keycloak-with-experimental-settings.yaml
```diff
@@ -29,3 +29,6 @@ spec:
           - configMap:
               name: test-config
               mountPath: /test-config
+          - secret:
+              name: keytabs
```
slaskawi, line 33: We probably need a small definition of an example Secret in this file (see how the ConfigMap is done).

slaskawi (changes requested):
Great work! Just one small comment and we're ready to get it in. 

BTW, could you please run `make code/gen` and include the generated changes in your PR?

deploy/examples/keycloak/keycloak-with-experimental-settings.yaml
soleblaze: Ok, I added a example secret

soleblaze:
> Great work! Just one small comment and we're ready to get it in.
> 
> BTW, could you please run `make code/gen` and include the generated changes in your PR?

I'm hitting an error when running make code/gen with operator-sdk v0.18.2 that also happens on master:

```
❯ make code/gen
operator-sdk generate k8s
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [keycloak:[v1alpha1], ]
F0223 11:06:02.502842 1396519 deepcopy.go:813] Hit an unsupported type invalid type for []invalid type
make: *** [Makefile:158: code/gen] Error 255
```

Is there a different version I should be using? (It doesn't look like generate k8s is available on the latest version)

soleblaze:
Looks like my issue was due to $GOROOT not being set. I've pushed the code/gen updates.

soleblaze:
I'm going to close this in favor of putting out another PR that also adds the ability to mount multiple configmaps/secrets to the same location.

kfox1111:
Can you please ping this thread when you do so I can follow it? :)

Thanks for working on this feature.

soleblaze:
Opened up #321

PR #320, opened 2021-03-03
KEYCLOAK-17303 ImagePullSecret Environment Variables

trevorloranger:
## JIRA ID
[KEYCLOAK-17303](https://issues.redhat.com/browse/KEYCLOAK-17303)

## Additional Information
See JIRA issue for more details - but at a high-level this PR provides configuration of `imagePullSecret` by additional environment variables. This is intended to complement the previously existing environment variables for [external images](https://github.com/keycloak/keycloak-operator/pull/165).

## Checklist:
- [x] Automated Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,14 +176,16 @@ NOTE: This functionality works only in OpenShift environment.
 
 All images used by the Operator might be controlled using dedicated Environmental Variables:
 
- | *Image*             | *Environment variable*          | *Default*                                                        |
- | ------------------- | ------------------------------- | ---------------------------------------------------------------- |
- | `Keycloak`          | `RELATED_IMAGE_KEYCLOAK`                | `quay.io/keycloak/keycloak:9.0.2`                                |
- | `RHSSO` for OpenJ9  | `RELATED_IMAGE_RHSSO_OPENJ9`            | `registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4-1`        |
- | `RHSSO` for OpenJDK | `RELATED_IMAGE_RHSSO_OPENJDK`           | `registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4-1`        |
- | Init container      | `RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER` | `quay.io/keycloak/keycloak-init-container:master`                |
- | Backup container    | `RELATED_IMAGE_RHMI_BACKUP_CONTAINER`   | `quay.io/integreatly/backup-container:1.0.16`                    |
- | Postgresql          | `RELATED_IMAGE_POSTGRESQL`              | `registry.redhat.io/rhel8/postgresql-10:1`                       |
+ | *Image*             | *Image Environment variable*            | *Image Default*                                                   | *imagePullSecret Environment variable*      |
+ | ------------------- | --------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------- |
+ | `Keycloak`          | `RELATED_IMAGE_KEYCLOAK`                | `quay.io/keycloak/keycloak:latest`                                | `KEYCLOAK_IMAGE_PULL_SECRET`                |
+ | `RHSSO` for OpenJ9  | `RELATED_IMAGE_RHSSO_OPENJ9`            | `registry.redhat.io/rh-sso-7/sso74-openj9-openshift-rhel8:7.4`    | N/A                                         |
+ | `RHSSO` for OpenJDK | `RELATED_IMAGE_RHSSO_OPENJDK`           | `registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4`           | N/A                                         |
+ | Init container      | `RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER` | `quay.io/keycloak/keycloak-init-container:master`                 | `KEYCLOAK_INIT_CONTAINER_IMAGE_PULL_SECRET` |
+ | Backup container    | `RELATED_IMAGE_RHMI_BACKUP_CONTAINER`   | `quay.io/integreatly/backup-container:1.0.16`                     | `RHMI_BACKUP_CONTAINER_IMAGE_PULL_SECRET`   |
+ | Postgresql          | `RELATED_IMAGE_POSTGRESQL`              | `registry.access.redhat.com/rhscl/postgresql-10-rhel7:1`          | `POSTGRESQL_IMAGE_PULL_SECRET`              |
+
+*Note*: The `imagePullSecret` Environmental Variables allow you to specify a secret to be used when pulling the respective images.
 
 ## Contributing
 
--- a/pkg/model/image_manager.go
+++ b/pkg/model/image_manager.go
@@ -3,6 +3,8 @@ package model
 import (
 	"os"
 	"runtime"
+
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -175,14 +175,16 @@ NOTE: This functionality works only in OpenShift environment.
...
- | Init container      | `RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER` | `quay.io/keycloak/keycloak-init-container:master`                |
- | Backup container    | `RELATED_IMAGE_RHMI_BACKUP_CONTAINER`   | `quay.io/integreatly/backup-container:1.0.16`                    |
- | Postgresql          | `RELATED_IMAGE_POSTGRESQL`              | `registry.redhat.io/rhel8/postgresql-10:1`                       |
+ | *Image*             | *Image Environment variable*            | *Image Default*                                                   | *imagePullSecret Environment variable*      |
+ | ------------------- | --------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------- |
+ | `Keycloak`          | `RELATED_IMAGE_KEYCLOAK`                | `quay.io/keycloak/keycloak:latest`                                | `KEYCLOAK_IMAGE_PULL_SECRET`                |
+ | `RHSSO` for OpenJ9  | `RELATED_IMAGE_RHSSO_OPENJ9`            | `registry.redhat.io/rh-sso-7/sso74-openj9-openshift-rhel8:7.4`    | N/A                                         |
```
trevorloranger, line 181: Avoided `RHSSO` environment variables for `imagePullSecrets` since deploying the Operator using [methods 2 or 3](https://github.com/keycloak/keycloak-operator/pull/165#issuecomment-617099229) probably allows for secret configuration within OLM or by some other means (I'm only guessing as I haven't touched OLM or OpenShift).

pkg/model/image_manager.go
```diff
@@ -22,49 +24,72 @@ const (
...
+		},
+		PostgresqlImage: {
+			Image:           getImage(PostgresqlImage, DefaultPostgresqlImage),
+			ImagePullSecret: getImagePullSecret(postgresqlImagePullSecret),
+		},
 	}
 	return ret
 }
 
-func (p *ImageManager) getImage(environmentalVariable string, defaultValue string) string {
```
trevorloranger, line 48: Didn't seem to be using pointer receiver so I took it out

pkg/model/image_manager.go
```diff
@@ -22,49 +24,72 @@ const (
...
 
 var Images = NewImageManager()
 
-type ImageManager struct {
-	Images map[string]string
+type ImageManager map[string]Image
+
+type Image struct {
+	Image           string
+	ImagePullSecret v1.LocalObjectReference
```
trevorloranger, line 40: Could have made this `string` as well but this seemed more convenient.

pkg/model/image_manager.go
```diff
@@ -22,49 +24,72 @@ const (
 	DefaultRHSSOInitContainer    = "registry.redhat.io/rh-sso-7-tech-preview/sso74-init-container-rhel8:7.4"
 	DefaultRHMIBackupContainer   = "quay.io/integreatly/backup-container:1.0.16"
 	DefaultPostgresqlImage       = "registry.access.redhat.com/rhscl/postgresql-10-rhel7:1"
+
+	keycloakImagePullSecret              = "KEYCLOAK_IMAGE_PULL_SECRET"
```
trevorloranger, line 28: These aren't exported from package to avoid confusion and are not used as keys in `Images` map. Users would get secret for Keycloak by doing `Images[KeycloakImage].ImagePullSecret`

pkg/model/util.go
```diff
@@ -174,3 +174,13 @@ func roleMatches(a v1alpha1.RoleRepresentation, b v1alpha1.RoleRepresentation) b
 	}
 	return a.Name == b.Name
 }
+
+func filterEmptyImagePullSecrets(secrets []v1.LocalObjectReference) []v1.LocalObjectReference {
```
trevorloranger, line 178: Added this to make sure we aren't sending empty secrets during k8s API calls - which is probably okay to delegate to k8s server for validation but this is cleaner. Effectively, this PR won't modify the over-the-wire k8s API calls when the user doesn't specify any `imagePullSecret` environment variables.

README.md
```diff
@@ -175,14 +175,16 @@ NOTE: This functionality works only in OpenShift environment.
...
- | `RHSSO` for OpenJDK | `RELATED_IMAGE_RHSSO_OPENJDK`           | `registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4-1`        |
- | Init container      | `RELATED_IMAGE_KEYCLOAK_INIT_CONTAINER` | `quay.io/keycloak/keycloak-init-container:master`                |
- | Backup container    | `RELATED_IMAGE_RHMI_BACKUP_CONTAINER`   | `quay.io/integreatly/backup-container:1.0.16`                    |
- | Postgresql          | `RELATED_IMAGE_POSTGRESQL`              | `registry.redhat.io/rhel8/postgresql-10:1`                       |
+ | *Image*             | *Image Environment variable*            | *Image Default*                                                   | *imagePullSecret Environment variable*      |
+ | ------------------- | --------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------- |
+ | `Keycloak`          | `RELATED_IMAGE_KEYCLOAK`                | `quay.io/keycloak/keycloak:latest`                                | `KEYCLOAK_IMAGE_PULL_SECRET`                |
```
trevorloranger, line 180: Updated docs for default images as well

slaskawi:
Thanks for the interest in this @trevorloranger!

Sadly, this has been discussed many times in the past (more information might be found in [KEYCLOAK-12409](https://issues.redhat.com/browse/KEYCLOAK-12409)) and the approach you provided is a no-go. 

Making the long story short - the Operator takes an opinionated take on the way you deploy Keycloak in your cluster. Opinionated here means, this provides you a way to extend your deployment in very well defined places, such as extensions. It also prevents you from customizing things too much - e.g. providing your own images.

The same reasoning applies to Pull Secrets. Since the Operator doesn't allow you to override the images, there's no reason to provide a way to supply Pull Secrets.

As a workaround, I can suggest to create a Pull Secret in the namespace you'd like to deploy Keycloak. You can also deploy Keycloak Operator manually and override Operator's environment variables to specify your own image. Note, that this means you're running your own, customized Operator installation. There's a good reason, why you can't do things like this with the OLM.

Nevertheless, we'd like to learn more about **why** are you trying to use a custom image? Perhaps there's something we can do (provide an extension point somewhere?) to satisfy your case without replacing our images with custom ones?

trevorloranger:
@slaskawi thank you for the detailed response! Myself, like @kfox1111, are trying to get more clarity on why this approach is a no-go given the externalize images [functionality](https://github.com/keycloak/keycloak-operator/pull/165). Hopefully, the following commentary will encourage more dialog to bring us all to a better shared understanding:

1. We are deploying the Keycloak Operator [manually](https://github.com/keycloak/keycloak-operator/pull/165#issuecomment-617099229) without OLM. If users can already specify `images` as envionrment variables [here](https://github.com/keycloak/keycloak-operator/blob/master/deploy/operator.yaml#L23), why not allow users to additionally specify a corresponding `imagePullSecrets` as well? One could argue this goes against the 12-factor's rules for [configuration](https://12factor.net/config) since this is likely to vary between deployments.
2. There seems to be a disconnect whether the `RELATED_IMAGE_*` environment variables are considered part of the Keycloak Operator's "public" API. You could argue, yes, since they are explicitly mentioned as configuration items within the [README](https://github.com/keycloak/keycloak-operator/blob/master/README.md#components-versions). On the other hand, from the [comments](https://github.com/keycloak/keycloak-operator/pull/165#issuecomment-613392546) and [documentation](https://docs.openshift.com/container-platform/4.7/operators/admin/olm-restricted-networks.html) I've read, it seems these environment variables are interpreted as "implementation details" for the sole purpose of OLM offline installations. Is this accurate?
3. I'm confused by the following:
    > The same reasoning applies to Pull Secrets. Since the Operator doesn't allow you to override the images, there's no reason to provide a way to supply Pull Secrets.

    Isn't this possible with the mechanism I described above? I'm guessing, but did you mean the _OLM doesn't allow you override the images_?
4. Regarding:
    > As a workaround, I can suggest to create a Pull Secret in the namespace you'd like to deploy Keycloak. You can also deploy Keycloak Operator manually and override Operator's environment variables to specify your own image. Note, that this means you're running your own, customized Operator installation. There's a good reason, why you can't do things like this with the OLM.

    This is exactly what we are trying to do: deploying the Keycloak Operator manually with custom images. Unfortunately, this functionality isn't fully usable when the specified images are in a private registry. This can be solved in multiple [ways](https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry):
    1. During cluster provisioning one can either [configure nodes to authenticate with private registries](https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry) or [cache images](https://kubernetes.io/docs/concepts/containers/images/#pre-pulled-images). If you are a [Service Infrastructure Admin](https://github.com/slaskawi/keycloak-community/blob/Operator-high-level-architecture/design/operator-architecture.asciidoc#use-cases) and using a pre-provisioned cluster this won't always be possible.
    2. [Specifying `imagePullSecrets` on a Pod](https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod) which is the objective of this pull request. Also, one can add `imagePullSecrets` to a Service Account which is our current workaround. Indeed, all of the pods started by the Operator use the default service account which we [patch](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-image-pull-secret-to-service-account) with the `imagePullSecrets`. In general, patching the default service account with `imagePullSecrets` is generally not [advisable](https://security.stackexchange.com/a/229442) and it would be preferable to specify the service account used by every pod through an environment variable. What do you think of this alternative approach?

5. For your question:
    > Nevertheless, we'd like to learn more about why are you trying to use a custom image? Perhaps there's something we can do (provide an extension point somewhere?) to satisfy your case without replacing our images with custom ones?

     We have a custom image which contains our theme and a CI/CD pipeline which builds an image whenever it detects a new upstream Keycloak image. We're aware of the approach of using JARs to deploy themes, which we intend to prototype, but we also have use cases which include non-Kubernetes based deployments of Keycloak where our custom images are convenient for consumers. Thus, to migrate entirely to this approach would necessitate us migrating to this approach for Docker Compose workflows which seem to [lack init containers](https://github.com/docker/compose/issues/6855) - which are leveraged in this scenario.

In closing, we completely understand your concerns surrounding [maintenance burden](https://github.com/keycloak/keycloak-operator/pull/165#issuecomment-618256163) and one must always take careful consideration when evaluating the inclusion of new functionality. Thank you once again and we are looking forward to your guidance. Cheers!

slaskawi:
@trevorloranger I appreciate your arguments. Could you please summarize your post and put it in [Keycloak DEV](https://groups.google.com/g/keycloak-dev)? I believe it might help if a larger group will evaluate this.

trevorloranger:
@slaskawi [Done!](https://groups.google.com/g/keycloak-dev/c/Wth-64cH6aw/m/T9psXZ2qBAAJ) Thanks for quick response!

PR #327, opened 2021-03-17
KEYCLOAK-17450: adding support for the Attributes field of the Realm CRD

oliverkane:
## KEYCLOAK-17450:
I would like to be able to set the frontendUrl property of a Realm via the CRD.  This PR is to do that.

## Additional Information
Without this field, some clients (eg: oidc-node) fail to resolve the proper frontend URL when using an oidc discovery document, since keycloak will infer the hostname based on the http request.  This property overrides it.

## Verification Steps

1.     Create a realm without frontendUrl.
2.     Set the value and apply.
3.     Verify that the realm was updated as expected.


oliverkane:
Ok, Well. This is driving me bonkers. Gonna dump this work and see if i can get the image to build for me locally.

PR #334, opened 2021-03-25
KEYCLOAK-17107 Configuable application context path

apeschel:
## JIRA ID

The Keycloak JIRA login process is horribly broken for me, and I cannot login to the ticket board. If I can get help with the login process, I would gladly use it to collaborate.

https://issues.redhat.com/browse/KEYCLOAK-17107

## Additional Information

This PR is an attempt to make the context path configurable. Keycloak has a default context path of "/auth", but it is fully possible and supported to use another path. Unfortunately, with this operator the path has been hard coded to "/auth" into literally every place that a path is used. This makes it impossible to use any other path when using the operator.

This PR is an attempt to make the context path configurable.

There's a lot of stuff that touches the context path in this repository, and I'm not familiar enough with the code base to figure out some of the stuff that needs to be changed. If a maintainer is willing to help provide some guidance, I would be willing to do most of the heavy lifting.

### DIFF
--- a/pkg/apis/keycloak/v1alpha1/keycloak_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloak_types.go
@@ -82,6 +82,8 @@ type KeycloakSpec struct {
 	// Specify PodAntiAffinity settings for Keycloak deployment in Multi AZ
 	// +optional
 	MultiAvailablityZones MultiAvailablityZonesConfig `json:"multiAvailablityZones,omitempty"`
+	// Context Path used by WildFly server. Default to "auth".
+	ContextPath *string `json:"contextPath"`
 }
 
 type DeploymentSpec struct {
--- a/pkg/common/client.go
+++ b/pkg/common/client.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	authURL = "auth/realms/master/protocol/openid-connect/token"
+	contextPath = "auth"
+	authURL = fmt.Sprintf("%s/realms/master/protocol/openid-connect/token", contextPath)
 )
 
 type Requester interface {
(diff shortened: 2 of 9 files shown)


apeschel:
At the moment, I'm just trying to track down every location that hard codes the context path, and extract it to variables. Even if the operator still only works with "/auth", that should at least make it easier to start converting the variables to configuration.

apeschel:
It looks like there was some monitoring and metrics changes that were added recently that added hard coded paths that are going to be harder than most to correct.

slaskawi:
I'm not very keen on merging this Pull Request. Keycloak Operator is an opinionated way of deploying Keycloak server(s) in Kubernetes and OpenShift environment. Being opinionated also means using all the default we use - including the context path.

If you really must change the context path, please do it outside of the Keycloak Operator by creating your Route or Ingress manually. Then you can apply whatever rewrites you wish.

Before closing this Pull Request, let me ask you about your use case for this? Perhaps, there's some other way we can help you?

apeschel:
@slaskawi We thought we had an existing endpoint that used /auth, which would have meant a collision. After further review, it turned out there was actually no overlapping endpoint, and we can use the default /auth endpoint provided by Keycloak.

That's a potential use case though -- someone who does have an existing /auth endpoint they cannot change.

Your proposed workaround doesn't work, unfortunately. I've tried setting up an ingress resource that would forward /keycloak to /auth on the Keycloak server via rewrites.

However, Keycloak returns paths with the context path in them, so if you have a rewrite on your ingress that does something like forward /keycloak -> /auth, all the URLs in the response will still have /auth in them, rather than /keycloak. This results in a half-functional mess that doesn't really work.

slaskawi:
Thanks for the explanation @apeschel !

To me, this is a wrong layer to address this problem. The separation should happen on the Ingress/Route level by assigning different hostnames for the services. 

Let me close this Pull Request than. If you wish to dive more into this thread, please open up a new conversation on Keycloak DEV forum.

PR #332, opened 2021-03-23
KEYCLOAK-16220 KEYCLOAK-15895 Operator communicates with KC using external URLs

vmuzikar:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-16220
https://issues.redhat.com/browse/KEYCLOAK-15895

## Additional Information
I'm addressing two issues in a single PR as I believe it makes sense. [KEYCLOAK-15895](https://issues.redhat.com/browse/KEYCLOAK-15895) was blocked by [KEYCLOAK-16220](https://issues.redhat.com/browse/KEYCLOAK-16220) and at the same time [KEYCLOAK-15895](https://issues.redhat.com/browse/KEYCLOAK-15895) verifies that [KEYCLOAK-16220](https://issues.redhat.com/browse/KEYCLOAK-16220) fixed the problem.

## Checklist:
- [X] Automated Tests
- [ ] [Documentation](https://github.com/keycloak/keycloak-documentation) changes if necessary
- [ ] (If required) Discussed on [Keycloak DEV](https://groups.google.com/forum/#!forum/keycloak-dev)

PR #335, opened 2021-03-25
KEYCLOAK-17675 Allow supplying a server certificate to use when connecting to keycloak

clive-jevons:
If the sso-x509-https-secret secret is used to provision a server cert for the keycloak to use
then this will be loaded and set in the http client used to communicate
with the Keycloak server. If the env var is not set, then the behaviour is as before, e.g. insecure.

## Verification Steps

Run the unit test included in the PR. 

`go test -timeout 30s -run ^TestClient_useKeycloakServerCertificate$ github.com/keycloak/keycloak-operator/pkg/common`

## Checklist:
- [X ] Automated Tests

### DIFF
--- a/pkg/common/client.go
+++ b/pkg/common/client.go
@@ -929,14 +929,8 @@ func (i *LocalConfigKeycloakFactory) AuthenticatedClient(kc v1alpha1.Keycloak) (
 	user := string(adminCreds.Data[model.AdminUsernameProperty])
 	pass := string(adminCreds.Data[model.AdminPasswordProperty])
 
-	sslCertsSecret, err := secretClient.CoreV1().Secrets(kc.Namespace).Get(context.TODO(), model.ServingCertSecretName, v12.GetOptions{})
-	var serverCert []byte
-	switch {
-	case err == nil:
-		serverCert = sslCertsSecret.Data["tls.crt"]
-	case k8sErrors.IsNotFound(err):
-		serverCert = nil
-	default:
+	serverCert, err := getKCServerCert(secretClient, kc)
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


slaskawi (dismissed):
Thanks for submitting the PR @clive-jevons ! I like this change. 

@abstractj @vmuzikar Would you like to review this before I merge it?

vmuzikar (commented):
@clive-jevons Thank you for your contribution, I like this improvement!

Just one comment. I noticed this conflicts with #332, specifically in the `defaultRequester` which now requires a server cert. `defaultRequester` is used at more places in #332:
https://github.com/keycloak/keycloak-operator/blob/57ab8c1478ee5ab7a23ce8a40ac9350b04373fa5/pkg/common/client.go#L926-L929

I wonder if we could somehow extract the logic for getting the cert secret from `AuthenticatedClient` into a separate function. This way `defaultRequester` signature could remain the same or at least `GetKeycloakURL` in #332 could call that function and pass the cert to `defaultRequester`.

clive-jevons:
@slaskawi @vmuzikar  thanks for the feedback 😊

The reason I'd included the code in the AuthenticatedClient method was to avoid duplicating creating a Secrets client. So here's what I'm thinking *could* work:

* pull the code out into a private method `getKCServerCert` which takes a Secrets client
* Call this from `AuthenticatedClient`
* In #332 create an internal method which gets the URL and *also* receives a Secrets client. This method can then be called from `AuthenticatedClient`
* The public `GetKeycloakURL` method in #332 can then create its own Secrets client and call the new internal method.

That way for each execution path only one Secrets client needs creating.

How does that sound?

vmuzikar:
@clive-jevons Ok, sounds good. Even though I'm not sure how much of a perf overhead is creating a secrets client (and it's probably not too much "Go way", introducing a bit complexity for a small perf gain), I'm ok with it. Let's do it like that! I'll update my PR if this gets merged first.

clive-jevons:
@vmuzikar Alright - I've pushed a change which should reflect what we've discussed here. Would be much obliged if you could take another look 😊 Thanks 😁

vmuzikar (approved):

clive-jevons:
@slaskawi could you please have another look and maybe approve / merge if possible? 😬 Thanks v.much in advance 😁

slaskawi (approved):

slaskawi:
Integrated, thanks @clive-jevons !

clive-jevons:
Thank you both very much @slaskawi @vmuzikar 😁

PR #340, opened 2021-03-31
Update to latest stable keycloak-init container

apeschel:
KEYCLOAK-17330

The keycloak-init-container image is currently pinned to the "master"
shared tag, which has not been updated in a year.

This commit updates the image tag to use the latest stable tag for the
container.

This container comes with a couple of benefits:

* Accurate URL parsing, the Python parsing has issues.
* Ability to use basic auth URLs for extensions.
* Much smaller image size.

## JIRA ID

[KEYCLOAK-17330](https://issues.redhat.com/browse/KEYCLOAK-17330)

## Additional Information

Allows the use of private extensions via the use of a basic auth URL.

## Verification Steps

1. Create a keycloak resource with an extension using a basic auth URL.
2. Check the keycloak-init container log to verify it downloaded the extension
3. Check the keycloak container to verify the extension was installed correctly

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/pkg/model/constants.go
+++ b/pkg/model/constants.go
@@ -44,7 +44,7 @@ const (
 	RouteLoadBalancingStrategy            = "source"
 	IngressDefaultHost                    = "keycloak.local"
 	PostgresqlBackupServiceAccountName    = "keycloak-operator"
-	KeycloakExtensionEnvVar               = "KEYCLOAK_EXTENSIONS"
+	KeycloakExtensionEnvVar               = "KEYCLOAK_EXTENSION"
 	KeycloakExtensionPath                 = "/opt/jboss/keycloak/standalone/deployments"
 	KeycloakExtensionsInitContainerPath   = "/opt/extensions"
 	RhssoExtensionPath                    = "/opt/eap/standalone/deployments"
--- a/pkg/model/image_manager.go
+++ b/pkg/model/image_manager.go
@@ -18,7 +18,7 @@ const (
 	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:latest"
 	DefaultRHSSOImageOpenJ9      = "registry.redhat.io/rh-sso-7/sso74-openj9-openshift-rhel8:7.4"
 	DefaultRHSSOImageOpenJDK     = "registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4"
-	DefaultKeycloakInitContainer = "quay.io/keycloak/keycloak-init-container:master"
+	DefaultKeycloakInitContainer = "quay.io/keycloak/keycloak-init-container:12.0.4"
 	DefaultRHSSOInitContainer    = "registry.redhat.io/rh-sso-7-tech-preview/sso74-init-container-rhel8:7.4"
 	DefaultRHMIBackupContainer   = "quay.io/integreatly/backup-container:1.0.16"
 	DefaultPostgresqlImage       = "registry.access.redhat.com/rhscl/postgresql-10-rhel7:1"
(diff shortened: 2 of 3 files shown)


apeschel:
So it turns out the Extensions env var was changed upstream. It's not clear why -- maybe a mistake?

https://github.com/keycloak/keycloak-containers/commit/da4b38c469077b78b46d8b18cf5f97ed8910f520#diff-51b1ab20d6e750a4464ec80bb5021a47be580d89ccc4cd25427ee812c5893347R44

This PR now updates the env var to match the upstream change.

pkg/model/image_manager.go
```diff
@@ -18,7 +18,7 @@ const (
 	DefaultKeycloakImage         = "quay.io/keycloak/keycloak:latest"
 	DefaultRHSSOImageOpenJ9      = "registry.redhat.io/rh-sso-7/sso74-openj9-openshift-rhel8:7.4"
 	DefaultRHSSOImageOpenJDK     = "registry.redhat.io/rh-sso-7/sso74-openshift-rhel8:7.4"
-	DefaultKeycloakInitContainer = "quay.io/keycloak/keycloak-init-container:master"
+	DefaultKeycloakInitContainer = "quay.io/keycloak/keycloak-init-container:12.0.4"
```
vmuzikar, line 21: I don't think we should hardcode the init container version, at least not in upstream/master. Shouldn't we rather fix the `master` tag?

pkg/model/constants.go
```diff
@@ -44,7 +44,7 @@ const (
 	RouteLoadBalancingStrategy            = "source"
 	IngressDefaultHost                    = "keycloak.local"
 	PostgresqlBackupServiceAccountName    = "keycloak-operator"
-	KeycloakExtensionEnvVar               = "KEYCLOAK_EXTENSIONS"
+	KeycloakExtensionEnvVar               = "KEYCLOAK_EXTENSION"
```
vmuzikar, line 47: I'm a bit concerned here about backward compatibility.

vmuzikar (commented):
@apeschel Thank you for your contribution. Added a few comments but let's wait for @slaskawi.

apeschel:
@vmuzikar Thanks for the quick response! I have some follow up to your points for when you're ready. Looking forward to hearing from @slaskawi !

pkg/model/image_manager.go
apeschel: I feel pinning a stable version rather than using a floating tag is a better for the stability of the keycloak-operator. If you're using a floating tag, you can have the functionality of a container change from under you at any point. You can run full integration tests, verify everything works, publish a stable release, and only after some indeterminate time have all your stable releases break because a new incompatible version of a dependency image is published.

In the specific case of the image defined here, if a new master tag image had been published in the last year, it would have caused keycloak-operator to immediately stop being able to load extensions in all cases, due to a breaking change that was introduced in this PR:

https://github.com/keycloak/keycloak-containers/commit/da4b38c469077b78b46d8b18cf5f97ed8910f520#diff-51b1ab20d6e750a4464ec80bb5021a47be580d89ccc4cd25427ee812c5893347L33-R44

By extension, if we fixed the master tag as you are suggesting, it would also immediately break every keycloak-operators.

For these reasons, I strongly suggest using the pinned stable version rather than the floating tag, and ending the use of the master tag for images in the keycloak-containers repository.

pkg/model/constants.go
apeschel: There was a breaking change introduced in the upstream image, which I have a strong suspicion was the result of an accident. This change just matches the upstream breaking change.

The alternative would be to fix the change upstream, which is doable as well, if that's preferred.

keycloak/keycloak-containers@da4b38c#diff-51b1ab20d6e750a4464ec80bb5021a47be580d89ccc4cd25427ee812c5893347L33-R44

apeschel:
Hi @slaskawi,

Just checking in on any updates here.

slaskawi:
Unfortunately, this is a wrong way to approach this problem, and sadly, this PR can not be merged.

Indeed, it seems we don't update the Init Container's version when doing the release. It goes without saying, this is a pretty major bug.

Nevertheless, it needs to be fixed in a different way. We use an automated release scripts that change the release version just before pushing a tag. The version replacement happens [here](https://github.com/keycloak/keycloak-operator/blob/master/set-version.sh). The proper fix should also replace the init container image coordinates.

@apeschel Would you be interested in reworking your PR for this?

apeschel:
@slaskawi Yes, I would like to help out -- I'll take some time this week to figure out what is going on with the version replacement, and start a new PR.

apeschel:
@slaskawi Any thoughts on which direction to go with the `KEYCLOAK_EXTENSIONS` vs `KEYCLOAK_EXTENSION` issue?

slaskawi:
> Any thoughts on which direction to go with the KEYCLOAK_EXTENSIONS vs KEYCLOAK_EXTENSION issue?

@apeschel The fix should go into the init container than.

PR #346, opened 2021-05-06
enable UMA from within the operator

kervel:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-18024

## Additional Information

Currently its not possible to control the userManagedAccessAllowed property on keycloak realm using the operator. We want to setup a keycloak realm using the operator with UMA enabled, so we would need this setting.

## Verification Steps

1. create a KeycloakRealm custom resources with spec.realm userManagedAccessAllowed set to true
2. wait for the resource to be created in the keycloak server
3. check if UMA allowed setting is enabled on the created realm

## Checklist:
- [ ] Automated Tests


kervel:
We have tested it and its work

PR #348, opened 2021-05-06
Keycloak Operator 13.0.0 Operatorhub sync

vmuzikar:
This is just a sync commit. Please merge it after this 2 PRs get in:
* https://github.com/operator-framework/community-operators/pull/3767
* https://github.com/operator-framework/community-operators/pull/3766

### DIFF
--- a/deploy/olm-catalog/keycloak-operator/13.0.0/keycloak-operator.v13.0.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/keycloak-operator/13.0.0/keycloak-operator.v13.0.0.clusterserviceversion.yaml
@@ -1,0 +1,355 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    capabilities: Deep Insights
+    categories: Security
+    certified: 'False'
+    containerImage: 'quay.io/keycloak/keycloak-operator:13.0.0'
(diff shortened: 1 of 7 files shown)


vmuzikar:
@slaskawi I did not. :( Do we expect we have inconsistencies between code and CRDs?

vmuzikar:
@slaskawi Luckily, there are no changes after running `make code/gen`.

slaskawi:
@vmuzikar In general - no but sometimes it happens that some changes go under the radar. That's why I always re-generate them before issuing a new release to the Operatorhub.

vmuzikar:
@slaskawi Thanks, that sounds like a good idea. :)

Maybe we could have some GH Action that would warn the contributor if it detects changes to the CRDs after running `make code/gen`.

slaskawi:
Sounds like a plan!

vmuzikar:
@slaskawi I think we're good to go. Could you please merge?

kfox1111:
Would it make sense to make a catalog / bundle for keycloak-operator so that syncing is no longer required?

slaskawi:
@kfox1111 This idea circulates back from time to time. But there are some projects that actually depend on our structure here. I'll talk with them if we could break this dependency somehow.

slaskawi (approved):

slaskawi:
Integrated, thanks @vmuzikar

PR #351, opened 2021-06-01
KEYCLOAK-18242 Keycloak Operator: database connection failure without…

xeras:
Fixes https://issues.redhat.com/browse/KEYCLOAK-18242

### DIFF
--- a/pkg/model/postgresql_service_endpoints.go
+++ b/pkg/model/postgresql_service_endpoints.go
@@ -15,7 +15,8 @@ func PostgresqlServiceEndpoints(cr *v1alpha1.Keycloak) *v1.Endpoints {
 			Name:      PostgresqlServiceName,
 			Namespace: cr.Namespace,
 			Labels: map[string]string{
-				"app": ApplicationName,
+				"app":       ApplicationName,
+				"component": PostgresqlDeploymentComponent,
 			},
 		},
 		Subsets: []v1.EndpointSubset{{
@@ -57,6 +58,7 @@ func PostgresqlServiceEndpointsReconciled(cr *v1alpha1.Keycloak, currentState *v
 	}
 
 	reconciled.Subsets[0].Ports = []v1.EndpointPort{{
+		Name:     PostgresqlPortName,
 		Port:     int32(portAsInt),
 		Protocol: "TCP",
 	}}
--- a/pkg/model/constants.go
+++ b/pkg/model/constants.go
@@ -14,6 +14,7 @@ const (
 	KeycloakMetricsRouteRewritePath      = "/auth/realms/master"
 	PostgresqlDeploymentComponent        = "database"
 	PostgresqlServiceName                = ApplicationName + "-postgresql"
+	PostgresqlPortName                   = PostgresqlServiceName + "-port"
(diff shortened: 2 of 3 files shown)


pkg/model/postgresql_service_endpoints.go
```diff
@@ -15,7 +15,8 @@ func PostgresqlServiceEndpoints(cr *v1alpha1.Keycloak) *v1.Endpoints {
 			Name:      PostgresqlServiceName,
 			Namespace: cr.Namespace,
 			Labels: map[string]string{
-				"app": ApplicationName,
+				"app":       ApplicationName,
+				"component": PostgresqlDeploymentComponent,
```
vmuzikar, line 19: The `component` label doesn't seem to be applied for some reason. I haven't investigated it deeper, though. Is it really working for you?

vmuzikar (changes requested):
@xeras Thank you for your contribution! In general LGTM, although the newly added label doesn't seem to work.

pkg/model/postgresql_service_endpoints.go
xeras: You are right, when it try to verify the changes with the following commands:
```
kubectl get svc -n keycloak-operator keycloak-postgresql -o yaml
kubectl get endpoints -n keycloak-operator keycloak-postgresql -o yaml
```
... the component label isn't listed. But without these changes, it doesn't work either.

So what do you suggest, should we remove the component label completely? I'm not sure, if this could be a problem.

vmuzikar: I believe this whole issue might be actually caused by [KEYCLOAK-18388](https://issues.redhat.com/browse/KEYCLOAK-18388). I think we might not need to name the ports after all, but I haven't tested that yet.

xeras: That is the same problem .... the service can't select the endpoint, because of the missing label. Removing the selector completely should result in the same fix. That's what i was asking for ... adding or removing :)

xeras: But i tried to just adding the selector label, without named ports ... that wasn't working. It just working when i fixed both issues.

vmuzikar: AFAICT, the real problem here is that there are any labels at all. We need to remove them (as described in KEYCLOAK-18388). I think that might do the trick and maybe naming the ports wouldn't be necessary.

Feel free to give it a try. :)

xeras: I can confirm that this just happens, if i want to connect the database by ip address instead of fqdn. When i provide a fqdn, the endpoint is not created ant this issue don't occurs. This works for me! *happy*

Also, i tried to remove the selector completely from the service definition, but then the endpoint wasn't created.
So i tried my previous solution, but i wasn't able to reproduce the fix. I tried providing a name for the ports, adding the missing label, removing the additional label, removing the whole selector .... nothing works today.

Because the fqdn works me, feel free to close this issue ... i don't have any new ideas how to fix that behavior. I'm really sorry about that :(

vmuzikar: Ok, thanks for letting me know. I'm happy you've found a solution! :)

PR #353, opened 2021-06-16
KEYCLOAK-18413 Upgrade deprecated APIs to ensure compatibility with K8s >=1.22

vmuzikar:


Rajagopalan-Ranganathan:
@vmuzikar - Can run `go mod vendor` and `go mod vendor`  to check if it affects `go.sum` file? When i run i locally in the branch go.sum has a few changes. Just to be sure!

vmuzikar:
I believe I've found the culprit – it was a bug after all. I forgot to update the Role with the new API group. I originally missed this as K8s on Docker for Mac is apparently acting differently from minikube.

@Rajagopalan-Ranganathan Thank you for the review! I ran `go mod vendor` and there were no changes whatsoever. What exactly is changed for you?

Rajagopalan-Ranganathan:
Have reviewed the changes that actually are there in the source code repo and not in the vendor/ path. It looks good.

abstractj (approved):
LGTM

PR #357, opened 2021-06-30
KEYCLOAK-17162: Add Support for ServiceAccount

christianwoehrle:
## JIRA ID 

KEYCLOAK-17162: Add Support for ServiceAccount

## Additional Information
The  Keycloak StatefulSet currently has the serviceAccountName set to be "default". 

This is sufficient in most cases, but we need to pull the Keycloak Image from a repository that requires an imagePullSecret.
This doesn't work with the "default" ServiceAccount.
 
As we deploy everything with helm charts, it is very difficult to change the existing "default"-ServiceAccount.

We would like to create a specific ServiceAccount for Keycloak that has the imagePullSecrets set as needed and configure Keycloak to use this Service Account.

This pull request enables this without changing the default behavior if the serviceAccountName is not set.

## Verification Steps

```
k apply -f -
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  extensions:
    - https://github.com/aerogear/keycloak-metrics-spi/releases/download/1.0.4/keycloak-metrics-spi-1.0.4.jar
  externalAccess:
    enabled: True
  podDisruptionBudget:
    enabled: True
```
```
kubectl get statefulset keycloak -oyaml | grep serviceAccountNa
serviceAccountName: default
```

-->

## Checklist:
- [ ] Automated Tests

### DIFF
--- a/deploy/crds/keycloak.org_keycloaks_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloaks_crd.yaml
@@ -876,6 +876,9 @@ spec:
                           - name
                           type: object
                         type: array
+                      serviceAccountName:
+                        description: ServiceAccountName settings
+                        type: string
                       volumes:
                         description: Additional volume mounts
                         properties:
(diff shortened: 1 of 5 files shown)


vmuzikar (approved):
@christianwoehrle Thank you for your contribution! Changes look good to me.

PR #363, opened 2021-07-18
KEYCLOAK-18868 Keycloak user secret namespace fix

Robbilie:
## Additional Information
Currently something like this can happen:
```
{"level":"info","ts":1626538984.2275712,"logger":"action_runner","msg":"(    2)     FAILED create credential secret for user realm_user in realm services/main"}
{"level":"debug","ts":1626538984.2278504,"logger":"controller-runtime.manager.events","msg":"Warning","object":{"kind":"KeycloakUser","namespace":"services","name":"example-user","uid":"de368f9a-d958-427b-bcf8-94d68389e479","apiVersion":"keycloak.org/v1alpha1","resourceVersion":"17349053"},"reason":"ProcessingError","message":"cross-namespace owner references are disallowed, owner's namespace services, obj's namespace keycloak"}
```
The realm is in the keycloak namespace, the user is in the services namespace though, thus not only services/main is incorrect but also a cross namespace resource cannot be created

## Verification Steps

1. Setup keycloak-operator to watch multiple namespaces (check operator sdk instructions)
2. Create a realm in one namespace
3. Create user in another namespace
4. Check operator logs (possibly on debug)

## Additional Notes 
Clients can already be created accross namespaces, the secret is properly created in the correct namespace so this here is just fixing it for users and bringing it in line.

### DIFF
--- a/pkg/controller/keycloakrealm/keycloakrealm_reconciler.go
+++ b/pkg/controller/keycloakrealm/keycloakrealm_reconciler.go
@@ -97,8 +97,12 @@ func (i *KeycloakRealmReconciler) getDesiredUserSate(state *common.RealmState, c
 	val, ok := state.RealmUserSecrets[user.UserName]
 	if !ok || val == nil {
 		return &common.GenericCreateAction{
-			Ref: model.RealmCredentialSecret(cr, user, &i.Keycloak),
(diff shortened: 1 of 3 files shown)


vmuzikar:
@Robbilie Thank you for your contribution! We currently do not support running the operator in multi namespace setups, therefore I'm going to close this.

PR #365, opened 2021-07-19
KEYCLOAK-18869 Authentication flow binding overrides

Robbilie:
## Additional Information
Currently you can create custom authentication flows on a realm CRD but you cannot use them in clients.

### DIFF
--- a/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakrealms_crd.yaml
@@ -265,6 +265,11 @@ spec:
                             type: string
                           description: Client Attributes.
                           type: object
+                        authenticationFlowBindingOverrides:
+                          additionalProperties:
+                            type: string
+                          description: Authentication Flow Binding Overrides.
+                          type: object
                         authorizationServicesEnabled:
                           description: True if fine-grained authorization support
                             is enabled for this client.
--- a/deploy/crds/keycloak.org_keycloakclients_crd.yaml
+++ b/deploy/crds/keycloak.org_keycloakclients_crd.yaml
@@ -47,6 +47,11 @@ spec:
                       type: string
                     description: Client Attributes.
                     type: object
+                  authenticationFlowBindingOverrides:
+                    additionalProperties:
+                      type: string
+                    description: Authentication Flow Binding Overrides.
+                    type: object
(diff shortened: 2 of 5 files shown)


Robbilie:
@vmuzikar @abstractj @stianst  how is the process to get something approved and merged? :)
are you occasionally checking PRs or do they have to be assigned to any of you?

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
```diff
@@ -265,6 +265,11 @@ spec:
                             type: string
                           description: Client Attributes.
                           type: object
+                        authenticationFlowBindingOverrides:
+                          additionalProperties:
+                            type: string
+                          description: Authentication Flow Binding Overrides.
+                          type: object
```
vmuzikar, line 272: This does not belong to the Realm CRD.

Please do not make manual changes to the CRDs and always run `make code/gen`. Thank you.

vmuzikar (changes requested):
@Robbilie Thank you for your contribution!

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
Robbilie: uhm i just reverted the yaml changes and ran the command and the result is identical to my changes…?

Robbilie: @vmuzikar can you verify this?

Robbilie: @vmuzikar can you verify that?

vmuzikar: Sorry again for the late response. I realized it is actually ok. The Client API representation is really included in the Realm CR as it is possible to define Clients there. Sorry about the confusion.

vmuzikar (approved):

deploy/crds/keycloak.org_keycloakrealms_crd.yaml
Robbilie: no worries, i wasnt following the proper steps in my first try so it couldve just as well been an error on my side so i verrified, so even though unintentionally, you made me follow proper procedures :P

PR #378, opened 2021-09-07
KEYCLOAK-19442 ssl db access

jonathanvila:
## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-19442

## Additional Information

To allow connection to PostgreSQL DB using TLS

## Verification Steps
1. install a Postgres DB with TLS only enabled using the Crunchy Operator 
  1. https://console-openshift-console.apps.playground.keycloak-team.engineering.redhat.com/operatorhub/ns/crossdc?keyword=crunch&details-item=crunchy-postgres-operator-certified-operators-openshift-marketplace
      1. use the yaml approach as using the form will result in errors
```
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  finalizers:
    - postgres-operator.crunchydata.com/finalizer
  name: example
  namespace: {your namespace}
spec:
  backups:
    pgbackrest:
      repos:
        - name: repo1
          volume:
            volumeClaimSpec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 1Gi
      restore:
        enabled: false
        repoName: repo1
  instances:
    - dataVolumeClaimSpec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
      name: ''
      replicas: 1
  openshift: true
  patroni:
    leaderLeaseDurationSeconds: 30
    port: 8008
    syncPeriodSeconds: 10
  port: 5432
  postgresVersion: 13
  proxy:
    pgBouncer:
      port: 5432
      replicas: 1
  standby:
    enabled: false
    repoName: repo1
```
  2. old approach = https://blog.crunchydata.com/blog/set-up-tls-for-postgresql-in-kubernetes 
2. copy values from the secret `example-pguser-example` (dbname, host, password, user)
3. create s secret ( `keycloak-db-secret` ) with these properties ( replace ? with the correct value )
POSTGRES_DATABASE=?
POSTGRES_EXTERNAL_ADDRESS=?
POSTGRES_EXTERNAL_PORT=?
POSTGRES_HOST=?
POSTGRES_PASSWORD=?
POSTGRES_USERNAME=?
SSLMODE=verify-ca
4. copy secret `pgo-root-cacert` to another secret called `keycloak-db-ssl-cert-secret`
3. execute the code in GoLand with the running conf ( https://github.com/keycloak/keycloak-operator/pull/383 ) paying attention to set the env var to the right image and setting the value for the namespace
```
WATCH_NAMESPACE={your namespace}
RELATED_IMAGE_KEYCLOAK=quay.io/keycloak/keycloak:15.0.2
RELATED_IMAGE_RHSSO=registry-proxy.engineering.redhat.com/rh-osbs/rh-sso-7-sso75-openshift-rhel8:rh-sso-7.5-rhel-8-containers-candidate-44208-20211026150954
```
image

4. create a Keycloak CR with ExternalDatabase=true and profile:RHSSO
```
kind: Keycloak
metadata:
  name: example-keycloak
  labels:
    app: sso
spec:
  instances: 1
  externalAccess:
    enabled: True
  podDisruptionBudget:
    enabled: True
  externalDatabase:
    enabled: True
  profile: RHSSO
```
5. after a while check that everything is running fine

## Checklist:
- [X] Automated Tests


pkg/model/keycloak_deployment.go
```diff
@@ -245,9 +245,39 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) *v13.Statefu
...
+					SecretName: DatabaseSecretSslCert,
+					Optional:   &[]bool{true}[0],
+				},
+			},
+		})
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, v1.VolumeMount{
+			Name:      DatabaseSecretSslCert + "_vol",
+			ReadOnly:  true,
+			MountPath: "/home/jboss/.postgresql",
+		})
```
vmuzikar, line 277: We do not need to always mount the certificate. E.g. in case of [`sslMode=allow`](https://jdbc.postgresql.org/documentation/head/ssl-client.html) the cert is not verified and therefore it doesn't need to be mounted.

I'd suggest we mount it only if the secret with the cert is present in the cluster.

pkg/model/constants.go
```diff
@@ -54,4 +54,7 @@ const (
 	MaxUnavailableNumberOfPods            = 1
 	ServiceMonitorName                    = ApplicationName + "-service-monitor"
 	MigrateBackupName                     = "migrate-backup"
+	DatabaseSecretSslModeProperty         = "SSLMODE"
+	DatabaseSecretSslCert                 = ApplicationName + "-db-ssl-cert-secret"
+	DatabaseConnectionParamsProperty      = "DB_CONN_PROPERTY"
```
vmuzikar, line 59: While this is valid for RH-SSO, please note that this property has a different name in Keycloak image (`JDBC_PARAMS`), see the [docs](https://github.com/keycloak/keycloak-containers/blob/master/server/README.md#specify-jdbc-parameters).

pkg/model/keycloak_deployment.go
```diff
@@ -245,9 +245,39 @@ func KeycloakDeployment(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) *v13.Statefu
...
+				Secret: &v1.SecretVolumeSource{
+					SecretName: DatabaseSecretSslCert,
+					Optional:   &[]bool{true}[0],
+				},
+			},
+		})
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, v1.VolumeMount{
+			Name:      DatabaseSecretSslCert + "_vol",
+			ReadOnly:  true,
+			MountPath: "/home/jboss/.postgresql",
```
vmuzikar, line 276: While this is valid for RH-SSO, please note that Keycloak image needs a different mount path: `/opt/jboss/.postgresql`

vmuzikar (commented):
Nice work so far, @jonathanvila! Thanks!

Even though it's still a draft, I added a few comments.

pkg/model/constants.go
jonathanvila: :+1:

pkg/model/keycloak_deployment.go
jonathanvila: :+1:

pkg/model/keycloak_deployment.go
jonathanvila: Very good point :+1:

jonathanvila: Considering the point you opened..... the same that applies to "allow" would apply to "disable" ? any special treatment should we do to "prefer" ? 
<p class="c3" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 1rem; color: rgb(51, 103, 145); font-family: &quot;Open Sans&quot;, sans-serif; font-size: 14.4px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 600; letter-spacing: normal; orphans: 2; text-align: left; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Table 31-1. SSL Mode Descriptions</p>

sslmode | Eavesdropping protection | MITM protection | Statement
-- | -- | -- | --
disable | No | No | I don't care about security, and I don't want to pay the overhead of encryption.
allow | Maybe | No | I don't care about security, but I will pay the overhead of encryption if the server insists on it.
prefer | Maybe | No | I don't care about encryption, but I wish to pay the overhead of encryption if the server supports it.
require | Yes | No | I want my data to be encrypted, and I accept the overhead. I trust that the network will make sure I always connect to the server I want.
verify-ca | Yes | Depends on CA-policy | I want my data encrypted, and I accept the overhead. I want to be sure that I connect to a server that I trust.
verify-full | Yes | Yes | I want my data encrypted, and I accept the overhead. I want to be sure that I connect to a server I trust, and that it's the one I specify.

vmuzikar: I would not implement any special behavior for the specific SSL modes in the operator. IMHO the operator doesn't need to understand the modes, doesn't need to care about their meaning – it just needs to pass it to the correct env var in the image.

I think we could keep it simple: if `sslMode` is specified AND the secret with the cert is found, we mount it. Don't need to based on the modes.

WDYT?

jonathanvila:
I prefer that approach. :+1

pkg/model/keycloak_deployment.go
jonathanvila: :+1:

test/e2e/keycloaks_test.go
```diff
@@ -42,6 +45,201 @@ func NewUnmanagedKeycloaksCRDTestStruct() *CRDTestStruct {
...
+	volumeExists := false
+	for _, vol := range keycloakStatefulset.Spec.Template.Spec.Volumes {
+		if vol.Name == model.DatabaseSecretSslCert+"-vol" {
+			volumeExists = true
+			break
+		}
+	}
+	if !volumeExists {
+		return errors.Errorf("test Failed : Volume to the secret not found")
+	}
```
vmuzikar, line 89: I believe you don't need to use a `for` loops which is a bit cumbersome. You can do something like [this](https://stackoverflow.com/a/2050629) instead. But it's just a nitpick – I'll leave it up to you if you want to change it. ;)

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -375,6 +376,109 @@ func TestKeycloakReconciler_Test_Updating_External_Database_URI(t *testing.T) {
...
+	}
+	assert.True(t, envVarOk)
+
+	sslVolumeExists := false
+	for _, volume := range keycloakSpec.Volumes {
+		if strings.Contains(volume.Name, model.DatabaseSecretSslCert+"-vol") {
+			sslVolumeExists = true
+		}
+	}
+	assert.True(t, sslVolumeExists)
```
vmuzikar, line 410: I believe you don't need to use a `for` loops which is a bit cumbersome. You can do something like [this](https://stackoverflow.com/a/2050629) instead. But it's just a nitpick – I'll leave it up to you if you want to change it. ;)

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -375,6 +376,109 @@ func TestKeycloakReconciler_Test_Updating_External_Database_URI(t *testing.T) {
...
+	}
+	assert.Equal(t, 2, envVarFound)
+
+	sslVolumeExists := false
+	for _, volume := range keycloakSpec.Volumes {
+		if strings.Contains(volume.Name, model.DatabaseSecretSslCert+"-vol") {
+			sslVolumeExists = true
+		}
+	}
+	assert.True(t, sslVolumeExists)
```
vmuzikar, line 447: I believe you don't need to use a `for` loops which is a bit cumbersome. You can do something like [this](https://stackoverflow.com/a/2050629) instead. But it's just a nitpick – I'll leave it up to you if you want to change it. ;)

pkg/controller/keycloak/keycloak_reconciler_test.go
```diff
@@ -375,6 +376,109 @@ func TestKeycloakReconciler_Test_Updating_External_Database_URI(t *testing.T) {
...
+	}
+	assert.False(t, sslVolumeExists)
+
+	envVarOk := false
+	for _, a := range keycloakSpec.Containers[0].Env {
+		if a.Name == model.KeycloakDatabaseConnectionParamsProperty && strings.Contains(a.Value, "sslmode") {
+			envVarOk = true
+		}
+	}
+	assert.False(t, envVarOk)
```
vmuzikar, line 479: I believe you don't need to use a `for` loops which is a bit cumbersome. You can do something like [this](https://stackoverflow.com/a/2050629) instead. But it's just a nitpick – I'll leave it up to you if you want to change it. ;)

pkg/model/keycloak_deployment.go
```diff
@@ -174,10 +174,35 @@ func getKeycloakEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
...
+
+		if sslMode != "" {
+			dbParams := ""
+			// is the deployment already having JDBC_PARAMS set ?
+			for _, element := range env {
+				if strings.EqualFold(element.Name, KeycloakDatabaseConnectionParamsProperty) {
+					dbParams = element.Value + "&"
+					break
+				}
+			}
```
vmuzikar, line 194: I believe you don't need to use a `for` loops which is a bit cumbersome. You can do something like [this](https://stackoverflow.com/a/2050629) instead. But it's just a nitpick – I'll leave it up to you if you want to change it. ;)

pkg/model/constants.go
```diff
@@ -31,27 +31,34 @@ const (
...
+	RhssoExtensionPath                         = "/opt/eap/standalone/deployments"
+	ClientSecretName                           = ApplicationName + "-client-secret"
+	ClientSecretClientIDProperty               = "CLIENT_ID"
+	ClientSecretClientSecretProperty           = "CLIENT_SECRET"
+	MaxUnavailableNumberOfPods                 = 1
+	ServiceMonitorName                         = ApplicationName + "-service-monitor"
+	MigrateBackupName                          = "migrate-backup"
+	DatabaseSecretSslModeProperty              = "SSLMODE"
+	DatabaseSecretSslCert                      = ApplicationName + "-db-ssl-cert-secret"
+	RhssoDatabaseXAConnectionParamsProperty    = "DB_XA_CONNECTION_PROPERTY"
```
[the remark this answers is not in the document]
jonathanvila: Who says that XA datasource has been removed ?

pkg/controller/keycloak/keycloak_reconciler_test.go
jonathanvila: The solution you linked is for maps, right ? In this case we don't have maps but a slice : []v1.Volume

pkg/model/keycloak_deployment.go
vmuzikar: Yes, you're of course correct. I did not realize it's not a map. Sorry for the confusion!

pkg/common/cluster_state.go
```diff
@@ -52,6 +52,7 @@ type ClusterState struct {
 	KeycloakPrometheusRule          *monitoringv1.PrometheusRule
 	KeycloakGrafanaDashboard        *grafanav1alpha1.GrafanaDashboard
 	DatabaseSecret                  *v1.Secret
+	DatabaseSSLSecret               *v1.Secret
```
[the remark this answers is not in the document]
vmuzikar: > This is a more generic comment, not only related to this line. I'd maybe rename `DatabaseSSLSecret` to something like `DatabaseSSLCert` to be more clear about the purpose of the Secret. WDYT?

I believe this ^^ is still valid.

pkg/model/constants.go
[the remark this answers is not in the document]
vmuzikar: > As far as I know there was a ticket to remove it, but it has been closed. Do you have more info about this removal ?

Is it still valid that we need `DB_XA_CONNECTION_PROPERTY`?

test/e2e/keycloaks_test.go
```diff
@@ -42,6 +45,201 @@ func NewUnmanagedKeycloaksCRDTestStruct() *CRDTestStruct {
...
+	// deploy PostgreSQL
+	postgresvc, err := deployPostgreSQLWithSSLon(secretWithSSLCertForPostgres, cr, secret, f, ctx)
+	if err != nil {
+		return err
+	}
+
+	// deploying the Keycloak CR
+	cr.Spec.ExternalDatabase.Enabled = true
+	secret.Data["SSLMODE"] = []byte("verify-ca")
+	secret.Data["POSTGRES_EXTERNAL_ADDRESS"] = []byte(postgresvc.Name + ".keycloak.svc.cluster.local")
```
vmuzikar, line 115:
```suggestion
	secret.Data["POSTGRES_EXTERNAL_ADDRESS"] = []byte(postgresvc.Name + "." + namespace + ".svc.cluster.local")
```

testdata/pg_hba.conf
```diff
@@ -0,0 +1,95 @@
+# PostgreSQL Client Authentication Configuration File
```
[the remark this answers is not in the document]
vmuzikar: > Do we need this file? I didn't find any mention of it being used anywhere.

I believe this is still valid.

pkg/common/cluster_state.go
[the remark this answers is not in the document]
jonathanvila: Ok

test/e2e/keycloaks_test.go
jonathanvila: :+1:

testdata/pg_hba.conf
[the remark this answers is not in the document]
jonathanvila: Looks that we don't need it, I used it before for the tests but no longer used.
:+1:

pkg/model/rhsso_deployment.go
```diff
@@ -121,10 +121,33 @@ func getRHSSOEnv(cr *v1alpha1.Keycloak, dbSecret *v1.Secret) []v1.EnvVar {
...
+
+		if sslMode != "" {
+			// append env variable
+			env = append(env,
+				v1.EnvVar{
+					Name:  RhssoDatabaseXAConnectionParamsProperty + "_sslMode",
+					Value: sslMode,
+				},
+				v1.EnvVar{
+					Name:  RhssoDatabaseNONXAConnectionParamsProperty + "_sslmode",
```
vmuzikar, line 141: Based on the discussion with @iankko, this needs to be:
```suggestion
					Name:  RhssoDatabaseNONXAConnectionParamsProperty + "_sslMode",
```

vmuzikar (commented):
Just to summarize, in order to merge this we need to do the following:
1. Update RH-SSO env var as stated below.
2. Test these changes with changes in RH-SSO image. Running the E2E test, that was added in this PR, is sufficient, no need to test manually.
3. Create docs PR for this change.

jonathanvila:
@vmuzikar documentation PR : https://github.com/keycloak/keycloak-documentation/pull/1299

vmuzikar (commented):
Just to update on the latest status. Changes LGTM, docs PR is ready. We now just need a confirmation this works with the RH-SSO image changes.

vmuzikar (approved):

PR #381, opened 2021-09-20
KEYCLOAK-19347 Keycloak Ingress doesn't work with Ingress Controller >= 1.0.0

vmuzikar:
* Added `kubernetes.io/ingress.class` annotation to make Ingress happy. Even though [some resources](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#deprecating-the-ingress-class-annotation) state that the annotation is deprecated, it still seem well supported (is even included in [official examples](https://kubernetes.github.io/ingress-nginx/user-guide/basic-usage/)) and it is much simpler and more straightforward solution than introducing [`IngressClass` object](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class).
* Upgraded minikube in the CI to use newer Ingress where the issue is reproducible.

### DIFF
--- a/pkg/model/keycloak_ingress.go
+++ b/pkg/model/keycloak_ingress.go
@@ -23,6 +23,7 @@ func KeycloakIngress(cr *kc.Keycloak) *networkingv1.Ingress {
 				"app": ApplicationName,
 			},
 			Annotations: map[string]string{
+				"kubernetes.io/ingress.class":                  "nginx",
 				"nginx.ingress.kubernetes.io/backend-protocol": "HTTPS",
 				"nginx.ingress.kubernetes.io/server-snippet": `
                       location ~* "^/auth/realms/master/metrics" {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,8 @@ jobs:
     - name: Start minikube
       uses: manusa/actions-setup-minikube@v2.4.2
       with:
-        minikube version: 'v1.21.0'
-        kubernetes version: 'v1.21.2'
+        minikube version: 'v1.23.1'
+        kubernetes version: 'v1.22.1'
         driver: 'docker'
     - name: Configure Minikube
       run: |


jonathanvila:
I am trying to reproduce the issue but I can't.
I am starting Minikube (1.23.1) with this ( 1.22.1 )  version of K8s, creating an ingress to a service ( without the annotation ) and I can't see the error : "Ignoring ingress because of error while validating ingress class" in the nginx-controller pod log

In order to see this is really solved, shouldnt we test it creating the ingress without the annotation, checking that the issue exists in the ingress-controller log, then add the annotation and check that the error no longer exists in the log ?

vmuzikar:
> I am trying to reproduce the issue but I can't.
> I am starting Minikube (1.23.1) with this ( 1.22.1 ) version of K8s, creating an ingress to a service ( without the annotation ) and I can't see the error : "Ignoring ingress because of error while validating ingress class" in the nginx-controller pod log
> 
> In order to see this is really solved, shouldnt we test it creating the ingress without the annotation, checking that the issue exists in the ingress-controller log, then add the annotation and check that the error no longer exists in the log ?

Thanks for the review. Minikube uses a beta version of Ingress 1.0.0 (while a stable version was already released). The log msg seems to be implementation specific and is not apparently present in the controller log in the beta version. The result is the same though – Ingress says 404.

pkg/model/keycloak_ingress.go
```diff
@@ -23,6 +23,7 @@ func KeycloakIngress(cr *kc.Keycloak) *networkingv1.Ingress {
 				"app": ApplicationName,
 			},
 			Annotations: map[string]string{
+				"kubernetes.io/ingress.class":                  "nginx",
```
jonathanvila, line 26: we are forcing the class of the ingress...... would this cause problems if the customer doesnt have nginx and has another controller installed ?
I think we should not enforce any specific class for ingress, storage, etc.

jonathanvila:
> > I am trying to reproduce the issue but I can't.
> > I am starting Minikube (1.23.1) with this ( 1.22.1 ) version of K8s, creating an ingress to a service ( without the annotation ) and I can't see the error : "Ignoring ingress because of error while validating ingress class" in the nginx-controller pod log
> > In order to see this is really solved, shouldnt we test it creating the ingress without the annotation, checking that the issue exists in the ingress-controller log, then add the annotation and check that the error no longer exists in the log ?
> 
> Thanks for the review. Minikube uses a beta version of Ingress 1.0.0 (while a stable version was already released). The log msg seems to be implementation specific and is not apparently present in the controller log in the beta version. The result is the same though – Ingress says 404.

I have run `master` branch of the Operator, on a minikube, and it works fine..... I still don't see where is the issue.
I have upgraded the NGINX ingress controller to v1 : 
https://kubernetes.github.io/ingress-nginx/deploy/upgrade/
https://github.com/kubernetes/ingress-nginx/tree/controller-v1.0.0

image
image

vmuzikar:
@jonathanvila As this may be apparently Ingress Controller issue, let's wait how https://github.com/kubernetes/ingress-nginx/issues/7674 develops.

vmuzikar:
@jonathanvila Thanks for the update! The operator already [assumes the Ingress to be Nginx](https://github.com/keycloak/keycloak-operator/blob/996c21bca9f1d948e784d4f1ef5caaba5088944e/pkg/model/keycloak_ingress.go#L26-L27). However, the user can and is expected to modify the Ingress manually (even e.g. the hostname) as the Ngnix specific annotations are [not overridden by the reconciliation loop](https://github.com/keycloak/keycloak-operator/blob/996c21bca9f1d948e784d4f1ef5caaba5088944e/pkg/model/keycloak_ingress.go#L61-L95). From this reason IMHO we can set the Ingress class (as we already assume Nginx) to make it working OOTB in most cases and we could perhaps document explicitly that people can modify the class. WDYT?

jonathanvila:
OK, is this constraint "operator assumes nginx controller" specified in the
documentation or is it a hidden constraint?

vmuzikar:
I'd say it's not a constraint, rather an assumption. And it's not documented AFAIK.

jonathanvila:
Well if you are installing something specifying an ingles class, that you
expect is there, I would say is a constraint, because if the user
doesnt have that ingress class, it wont work.

I would suggest either removing the specification of the ingress class, and
expect ( I think is totally understandable ) that the user has a default
ingress class, OR , specify in the documentation that the operator expects
that ingress class to be installed.

vmuzikar:
Ok, I'm closing this. The issue appears to manifest only with a misconfigured (without default ingress class) cluster in any way.

PR #382, opened 2021-09-23
[KEYCLOAK-18294] Implement Service Account Roles

ctron:
## JIRA ID

KEYCLOAK-18294

## Verification Steps

1. Create a new `KeycloakClient` resource, add `.spec.client.serviceAccountRoles`.
2. Verify they get synced to the service account of the client.

## Checklist:
- [ ] Automated Tests

## Additional Notes 

I had some issues building the operator in general. Updating the go.mod fixed the issue for me. The first two commits are about updating the build errors, and can later on be dropped if necessary.

The PR is an early draft and untested. Just as a starting point for a discussion.

### DIFF
--- a/pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
+++ b/pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
@@ -22,6 +22,9 @@ type KeycloakClientSpec struct {
 	// Scope Mappings
 	// +optional
 	ScopeMappings *MappingsRepresentation `json:"scopeMappings,omitempty"`
+	// Service account realm roles for this client.
+	// +optional
+	ServiceAccountRealmRoles []string `json:"serviceAccountRealmRoles,omitempty"`
 }
 
 // https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_mappingsrepresentation
--- a/pkg/common/client.go
+++ b/pkg/common/client.go
@@ -828,6 +828,52 @@ func (c *Client) Ping() error {
 	return nil
 }
 
+func (c *Client) GetServiceAccountUser(realmName, clientID string) (*v1alpha1.KeycloakAPIUser, error) {
+	result, err := c.get(fmt.Sprintf("%s/clients/%s/service-account-user", realmName, clientID), "service-account-user", func(body []byte) (T, error) {
+		user := &v1alpha1.KeycloakAPIUser{}
+		err := json.Unmarshal(body, user)
+		return user, err
+	})
+	if err != nil {
+		return nil, err
+	}
+	if result == nil {
+		return nil, nil
+	}
+	ret := result.(*v1alpha1.KeycloakAPIUser)
+	return ret, err
+}
+
+func (c *Client) SetServiceAccountRealmRoles(realmName, clientID string, roles []string) error {
+	// get the service account user
+	user, err := c.GetServiceAccountUser(realmName, clientID)
+	if err != nil {
+		return err
+	}
+	if user == nil {
+		return errors.Errorf("client does not have a service account user")
+	}
+
+	// set the realm roles
+	user.RealmRoles = roles
+
+	// update the user
+	return c.UpdateUser(user, realmName)
+}
+
+func (c *Client) GetServiceAccountRealmRoles(realmName, clientID string) ([]string, error) {
+	// get the service account user
+	user, err := c.GetServiceAccountUser(realmName, clientID)
+	if err != nil {
+		return nil, err
+	}
+	if user == nil {
+		return nil, errors.Errorf("client does not have a service account user")
+	}
+
+	return user.RealmRoles, nil
+}
+
 // login requests a new auth token from Keycloak
 func (c *Client) login(user, pass string) error {
 	form := url.Values{}
@@ -977,7 +1023,15 @@ type KeycloakInterface interface {
 	GetAuthenticatorConfig(configID, realmName string) (*v1alpha1.AuthenticatorConfig, error)
 	UpdateAuthenticatorConfig(authenticatorConfig *v1alpha1.AuthenticatorConfig, realmName string) error
 	DeleteAuthenticatorConfig(configID, realmName string) error
+
+	GetServiceAccountUser(realmName, clientID string) (*v1alpha1.KeycloakAPIUser, error)
+
+	SetServiceAccountRealmRoles(realmName, clientID string, roles []string) error
+	GetServiceAccountRealmRoles(realmName, clientID string) ([]string, error)
 }
+
+// check if Client implements KeycloakInterface
+var _ KeycloakInterface = &Client{}
 
 //go:generate moq -out keycloakClientFactory_moq.go . KeycloakClientFactory
 
--- a/pkg/common/cluster_actions.go
+++ b/pkg/common/cluster_actions.go
@@ -47,8 +47,12 @@ type ActionRunner interface {
 	RemoveRealmRole(obj *v1alpha1.KeycloakUserRole, userID, realm string) error
 	AssignClientRole(obj *v1alpha1.KeycloakUserRole, clientID, userID, realm string) error
 	RemoveClientRole(obj *v1alpha1.KeycloakUserRole, clientID, userID, realm string) error
+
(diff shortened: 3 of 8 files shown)


pkg/common/client_state.go
```diff
@@ -96,6 +96,7 @@ func (i *ClientState) readClientScopes(cr *kc.KeycloakClient, realmClient Keyclo
 	if err != nil {
 		return err
 	}
+
```
vmuzikar, line 99: I believe this change is unnecessary.

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
```diff
@@ -164,6 +164,9 @@ type KeycloakAPIClient struct {
 	// Authorization settings for this resource server.
 	// +optional
 	AuthorizationSettings *KeycloakResourceServer `json:"authorizationSettings,omitempty"`
+	// Service account roles for this client.
+	// +optional
+	ServiceAccountRoles []string `json:"serviceAccountRoles,omitempty"`
```
vmuzikar, line 169: I would not place this in `KeycloakAPIClient` as it represents the Client in the Keycloak REST API. And the REST API doesn't have any `serviceAccountRoles` field. I think it should be placed in `KeycloakClientSpec`.

pkg/common/client.go
```diff
@@ -799,6 +799,25 @@ func (c *Client) Ping() error {
 	return nil
 }
 
+// Get the service account name from a client
+func serviceAccountName(clientName string) string {
+	return "service-account-" + clientName
```
vmuzikar, line 804: This should not be hardcoded. Please try to get the username from [REST API](https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_getserviceaccountuser).

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
vmuzikar, line 169: Is this Realm Roles or Client Roles? I think we should clearly differentiate this, similarly to [what we have for users](https://github.com/keycloak/keycloak-operator/blob/537d3ea9fb71570a4b7eb51fb6460a28b309450c/pkg/apis/keycloak/v1alpha1/keycloakuser_types.go#L71-L76) (because service account _is_ a user).

pkg/common/cluster_actions.go
```diff
@@ -557,6 +566,14 @@ type RemoveClientRoleAction struct {
 	Msg      string
 }
 
+type SetServiceAccountRolesAction struct {
+	ClientName string
+	Ref        *v1alpha1.KeycloakClient
+	Roles      []string
+	Realm      string
+	Msg        string
+}
```
vmuzikar, line 575: Can't we perhaps use [existing actions](https://github.com/keycloak/keycloak-operator/blob/ac6619083c859e5843b7fc4d76e100d14b3d096f/pkg/common/cluster_actions.go#L530-L558)?

vmuzikar (changes requested):
@ctron Thank you for your contribution! I think this is moving in the right direction. Added a few comments. Additionally to that, I'd like to avoid the changes in dependencies if possible. I think adding service accounts support shouldn't require it. Also, we'll need tests for this (ideally some simple E2E test if possible).

pkg/common/cluster_actions.go
ctron: I thought about that too. However, we need to:
* get the user (by name)
* add set the roles
* update the users

Now, we also need to evaluate the name of the user as you suggested. That is a composite action, and I am not sure the code will be simpler if is will be split into different (existing) actions, passing along state.

So I modeled it into a single action, for the client to implement.

ctron:
I updated the PR, implementing the changed requested.

>  Additionally to that, I'd like to avoid the changes in dependencies if possible. I think adding service accounts support shouldn't require it.

Indeed. However I wasn't able to compile locally without that.

> Also, we'll need tests for this (ideally some simple E2E test if possible).

Absolutely. I just wanted to see if this heads in the right direction, and invest the work in that later.

ctron:
I updated the PR once more, using the "service account user" API you mentioned.

ctron:
Still, I have no idea why suddenly it thinks the "logrus" dependency is missing … while it is just there … :shrug:

vmuzikar (commented):
@ctron Thanks for the update! New changes look good to me.

Just to summarize what needs to be done in order to merge this:
* Add E2E test.
* Fix broken build and tests.
* Preferably remove the changes in the dependencies. Of course, unless we figure out why they are necessary.

Worth to mention is also this PR currently adds support just for realm roles, not client roles. Naturally, the ideal would be to have support for both to be consistent with standard users, but I don't see this as a hard requirement and blocker for this PR..

go.mod
```diff
@@ -25,6 +25,8 @@ require (
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible
 	github.com/go-logr/logr => github.com/go-logr/logr v0.1.0
+	github.com/mikefarah/yaml/v2 => gopkg.in/yaml.v2 v2.4.0
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad
```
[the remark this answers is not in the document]
vmuzikar: I ran:
```
$ go clean --modcache
$ go mod vendor
$ make code/compile
$ go fmt
```
Everything works for me with your branch. GH Actions seem to be also able to build the project. Not sure what's happening.  What go version do you run? Mine:
```
$ go version 
go version go1.17.1 darwin/amd64
```

vmuzikar (changes requested):
@ctron Could you also please check the failing test and linter?

go.mod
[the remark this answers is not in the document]
ctron: Ok, I think I know what is going on. I have `GOPROXY=direct` set, which ignores the Google cache and directly downloads dependencies. When I unset this, it downloads from the Google cache, which seems to serve a different version. So "SECURITY ERROR" seems true. I will enable the "go proxy" for this project.

vmuzikar (changes requested):
@ctron I am sorry for a very late response and thank you for the update.

Changes look good to me now but we need the tests in the CI to pass. Could you please look into that? Let me know if you need any help with it.

vmuzikar:
@ctron Hello, just wanted to ask if this is something you'd like to work on. As it appears, you were not the only one interested in this feature. :)

manu11th:
@ctron I am also interested in this feature so I would really appreciate if you could finish your work here. Thanks in advance!

pkg/apis/keycloak/v1alpha1/keycloakclient_types.go
```diff
@@ -22,6 +22,12 @@ type KeycloakClientSpec struct {
 	// Scope Mappings
 	// +optional
 	ScopeMappings *MappingsRepresentation `json:"scopeMappings,omitempty"`
+	// Service account realm roles for this client.
+	// +optional
+	ServiceAccountRealmRoles []string `json:"serviceAccountRealmRoles,omitempty"`
+	// Service account client roles for this client.
+	// +optional
+	ServiceAccountClientRoles []string `json:"serviceAccountClientRoles,omitempty"`
```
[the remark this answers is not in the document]
Tanakan: @vmuzikar 
`it is possible to assign any client role to service account (similarly to [standard users](https://github.com/keycloak/keycloak-operator/blob/9cd42c20c1d357e09af6d92648979b03b424ece8/pkg/apis/keycloak/v1alpha1/keycloakuser_types.go#L74-L76)).`

I don't understand the comments yet so please let me know.
Is my understanding correct?
- Currently (main branch), CRD(KeycloakUser, KeycloakClient) cannot associate a ClientRole with a ServiceAccount.
- If the user ID of the service account is known in advance, the ClientRole can be associated service-account using KeycloakUser.

example
```
spec:
  realmSelector:
    matchLabels:
      app: sso
  user:
    enabled: true
    emailVerified: false
    realmRoles:
      - default-roles-test
    clientRoles:
      realm-management:
        - manage-clients
    username: service-account-<ClientID>
    id: <service-account-userID>
```

vmuzikar:
Superseded by #545

PR #383, opened 2021-10-11
Keycloak 18388 remove selector with external DB

jonathanvila:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-18388

## Verification Steps

Installation
1. start minikube ( max 1.18.0 with ingress controller 0.40 )
2. make cluster/clean
3. make cluster/prepare
4. install a PostgreSQL instance ( operator Crunchy DB or other way ) 
5. create a Secret in the keycloak namespace with these properties and the correspondent values : 
POSTGRES_DATABASE
POSTGRES_EXTERNAL_ADDRESS
POSTGRES_EXTERNAL_PORT
POSTGRES_HOST
POSTGRES_PASSWORD
POSTGRES_USERNAME
SSLMODE
4. on GoLand, run this configuration :
image
5. Config
  5.1 WATCH_NAMESPACE=keycloak
  5.2 RELATED_IMAGE_KEYCLOAK=quay.io/keycloak/keycloak:15.0.2
6. This will use the k8s context you are using ( minikube ) and run the operator
5. Create a Keycloak CR with ExternalDatabase=true and 

Test
1. once everything is deployed and running, you can use the Keycloak dashboard
2. in the keycloak-postgresql service you see no Selector

## Checklist:
- [X] Automated Tests

### DIFF
--- a/pkg/model/postgresql_service.go
+++ b/pkg/model/postgresql_service.go
@@ -14,19 +14,21 @@ func getSpec(dbSecret *v1.Secret, serviceTypeExternal bool) v1.ServiceSpec {
 	spec := v1.ServiceSpec{}
 	isIPAddress := dbSecret != nil && dbSecret.Data[DatabaseSecretExternalAddressProperty] != nil && IsIP(dbSecret.Data[DatabaseSecretExternalAddressProperty])
 
-	if serviceTypeExternal && !isIPAddress {
-		spec.Type = v1.ServiceTypeExternalName
+	if serviceTypeExternal {
+		// KEYCLOAK-18388 no sense for selectors when we have an external DB
 		spec.Selector = nil
-		spec.ExternalName = GetExternalDatabaseHost(dbSecret)
+		// ExternalName doesn't work with IP addresses
+		if !isIPAddress {
+			spec.Type = v1.ServiceTypeExternalName
+			spec.ExternalName = GetExternalDatabaseHost(dbSecret)
+		} else {
+			spec.Type = v1.ServiceTypeClusterIP
+		}
 	} else {
 		spec.Type = v1.ServiceTypeClusterIP
-		if !isIPAddress {
-			spec.Selector = map[string]string{
-				"app":       ApplicationName,
-				"component": PostgresqlDeploymentComponent,
-			}
-		} else {
-			spec.Selector = nil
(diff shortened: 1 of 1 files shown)


pkg/model/postgresql_service.go
```diff
@@ -12,16 +12,21 @@ import (
 
 func getSpec(dbSecret *v1.Secret, serviceTypeExternal bool) v1.ServiceSpec {
 	spec := v1.ServiceSpec{}
+	isIPAddress := dbSecret != nil && dbSecret.Data[DatabaseSecretExternalAddressProperty] != nil && IsIP(dbSecret.Data[DatabaseSecretExternalAddressProperty])
```
vmuzikar, line 15: Maybe I'd just add a new param to this function instead of calculating `isIPAddress` here. We already know this in `keycloak_reconciler`. But it's just a nitpick, fix it only if you want. :)

pkg/model/postgresql_service.go
```diff
@@ -12,16 +12,21 @@ import (
...
+	if serviceTypeExternal && !isIPAddress {
 		spec.Type = v1.ServiceTypeExternalName
 		spec.Selector = nil
 		spec.ExternalName = GetExternalDatabaseHost(dbSecret)
 	} else {
 		spec.Type = v1.ServiceTypeClusterIP
-		spec.Selector = map[string]string{
-			"app":       ApplicationName,
-			"component": PostgresqlDeploymentComponent,
+		if !isIPAddress {
```
vmuzikar, line 23: Please just leave a short comment here explaining why are we not setting selectors for IP addresses (you can mention the JIRA ID).

vmuzikar (commented):
@jonathanvila Thank you! In general looks good to me. Even though it's still a draft, I added a few comments.

pkg/model/postgresql_service.go
jonathanvila: The idea of putting this inside the method and no adding extra args is that it can do the functionality without extra params, so it has all the information. If we add another param we , by design, could have inconsistency.

pkg/model/postgresql_service.go
jonathanvila: :+1:

vmuzikar (approved):
LGTM, thank you!

PR #385, opened 2021-10-18
KEYCLOAK-19405 add labels to pods from constants and env vars

jonathanvila:
Signed-off-by: Jonathan Vila <jvilalop@redhat.com>

## JIRA ID

https://issues.redhat.com/browse/KEYCLOAK-19405

## Additional Information

To allow extracting metrics based on labels

1. adapt the ns in the Makefile
2. change the values in the constants enum
3. make cluster/prepare
4. run the operator in the IDE
5. edit the keycloak.yaml example
`  keycloakDeploymentSpec:
    experimental:
      env:
        - name: PODLABEL_redhat.any.label.first
          value: "redhat-this-is-a-value"
        - name: PODLABEL_redhat.any.label.second
          value: "redhat-this-is-another-value"`
6. apply the keycloak.yaml
7. check the pod created , verify the labels are the sum of these ones plus the ones in the constants file

## Checklist:
- [X] Automated Tests


pkg/model/util_test.go
```diff
@@ -154,3 +154,33 @@ func TestKeycloakClientReconciler_Test_ClientScope_DifferenceIntersection(t *tes
...
+		"FirstLabel":  "firstLabelValue",
+		"SecondLabel": "secondLabelValue",
+	}
+
+	labels := map[string]string{
+		"app":       ApplicationName,
+		"component": KeycloakDeploymentComponent,
+	}
+	totalLabels := AddPodLabels(&cr, labels)
+	assert.Equal(t, 6, len(totalLabels))
```
vmuzikar, line 179: I realized we cannot hardcode this as the "constant" with default labels will be different in prod. We need to reflect the constant it this test.

vmuzikar (changes requested):

pkg/model/util_test.go
jonathanvila: You mean in "keycloak-operator-prod" ? You suggest to extract that 6 into a constant that will be modified in the "keycloak-operator-prod" ? @vmuzikar

vmuzikar: Yes. In prod it might be 6 or 7 or 8 or ... theoretically any number of labels. It depends on how we modify the [`PodLabels`](https://github.com/keycloak/keycloak-operator/blob/ccf70429d9cff8e4e6036dd0163564a13a5d0c32/pkg/model/constants.go#L59) constant in prod.

pkg/model/constants.go
```diff
@@ -55,3 +55,5 @@ const (
 	ServiceMonitorName                    = ApplicationName + "-service-monitor"
 	MigrateBackupName                     = "migrate-backup"
 )
+
+var PodLabels = map[string]string{}
```
vmuzikar, line 59: I just noticed that this is actually not marked as a constant (as the constants above). Is there any reason for it?

jonathanvila: A Hashmap can not be a constant.

pkg/model/util_test.go
jonathanvila: Ahhhh but take into account the test is NOT using what could be defined in the constants.go.
Few lines above the test is initializing the map : 
```
	PodLabels = map[string]string{
		"FirstLabel":  "firstLabelValue",
		"SecondLabel": "secondLabelValue",
	}

	labels := map[string]string{
		"app":       ApplicationName,
		"component": KeycloakDeploymentComponent,
	}
	totalLabels := AddPodLabels(&cr, labels)
	assert.Equal(t, 6, len(totalLabels))
```

vmuzikar: The test is using `AddPodLabels` which is using the constant:
https://github.com/keycloak/keycloak-operator/blob/ccf70429d9cff8e4e6036dd0163564a13a5d0c32/pkg/model/util.go#L224-L227

jonathanvila:
Yes and AddPodLabels is using PodLabels, if you pay attention :
1. if it is not using the one few lines above the test wont pass
2. the PodLabels assignation few lines above is using the Global var

pkg/model/util_test.go
jonathanvila: Yes and AddPodLabels is using PodLabels, if you a closer look :
1. if it is not using the one few lines above the test won't pass
2. the PodLabels assignation few lines above is using the Global var

So the AddPodLabels is using the global var PodLabels that is initialized in the same test.

vmuzikar: Ah, sorry, you're right. I somehow managed to miss that you are assigning a new value to the `PodLabels` "constant" a few lines above.

vmuzikar (changes requested):
@jonathanvila Changes LGTM, thank you!

Could you please just rebase to resolve conflicts? Also the author of the commit seems to be broken, not sure what happened there.

jonathanvila:
:+1:

vmuzikar:
@jonathanvila Not sure why but I still see a weird commit there (at least author wise)
Screenshot 2021-11-09 at 17 01 40
:

jonathanvila:
> @jonathanvila Not sure why but I still see a weird commit there (at least author wise): Screenshot 2021-11-09 at 17 01 40

To be honest I dont know, probably I did something wrong merging from main, and I added the change in file README from mcanevet

jonathanvila:
> > @jonathanvila Not sure why but I still see a weird commit there (at least author wise): Screenshot 2021-11-09 at 17 01 40
> 
> To be honest I dont know, probably I did something wrong merging from main, and I added the change in file README from mcanevet

But in any case is already in `main`
image

jonathanvila:
@vmuzikar is there anything missing in this PR or we can progress it ?

vmuzikar:
@jonathanvila Could you please try to rebase and squash the commits? I'd like the authors of the commit to be set properly. :)

vmuzikar (approved):

PR #406, opened 2021-11-15
[minor] Check generated code in CI

andreaTP:
## Additional Information
Setting up the tools for compiling this repository I found out that some generated sources have(probably) been formatted with a newer version of `gofmt`.
A few tweaks enable people to use `golang` coming from `asdf` as well, instead of installing globally Go 1.13.

## Verification Steps

The steps to verify that the generated code doesn't change the diff is implemented as a CI step

## Checklist:
- [X] Automated Tests


check-diff.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -ex
+
+if [ -n "$(git status --untracked-files=no --porcelain)" ]; then
+  echo "code needs to be regenerated: use \"make setup\" and check in any resulting diffs"
+  git --no-pager diff
+  exit 1
+fi
+
```
vmuzikar, line 8: I'd use a different name for this file. `check-diff` is too generic.

vmuzikar (changes requested):
@andreaTP Thanks, this should have been done a long time ago. :)

Please create a GH Issue for this.

Makefile
```diff
@@ -149,7 +149,7 @@ setup/mod/verify:
 .PHONY: setup/operator-sdk
 setup/operator-sdk:
 	@echo Installing Operator SDK
-	@curl -Lo operator-sdk ${OPERATOR_SDK_DOWNLOAD_URL} && chmod +x operator-sdk && sudo mv operator-sdk /usr/local/bin/
+	/usr/local/bin/operator-sdk version || (curl -Lo operator-sdk ${OPERATOR_SDK_DOWNLOAD_URL} && chmod +x operator-sdk && mv operator-sdk /usr/local/bin/)
```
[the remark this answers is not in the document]
andreaTP: it is, indeed, but is bad practice to do `sudo` operations in a `Makefile`, also, as soon as you assign correct permissions once, you don't need `sudo` anymore.

check-diff.sh
andreaTP: any proposal?

Issue #432, opened 2021-12-21
Make the e2e tests pass without depending on external extension


jonathanvila:
### Describe the bug

Currently the e2e tests are failing on the extension test because the keycloak-metrics-spi has an incompatiblity with KC 16.0.0
(https://github.com/aerogear/keycloak-metrics-spi/issues/124)

### Version

16.0.0

### Expected behavior

not failing

### Actual behavior

failing

### How to Reproduce?

add the aerogear/keycloak-metrics-spi extension ( by default in the CR )

PR #433, opened 2021-12-21
#432 make extension test not dependent on external spi

jonathanvila:
## Related GH Issue
#432 

## Checklist:
- [ ] Automated Tests


vmuzikar (changes requested):
@jonathanvila Thanks for the  fix!

The ideal solution would be a custom Keycloak extension just for testing purposes (and we'll have to do it for the new operator) but in this case it's IMHO ok to keep this approach. However, I'd rather rely on the [official 2.5.3 release](https://github.com/aerogear/keycloak-metrics-spi/releases/tag/2.5.3), not to hardcode any custom fork etc.

vmuzikar:
@jonathanvila bump :)

vmuzikar (approved):

PR #443, opened 2022-01-11
Allow Operator to Create / Delete Realms on unmanaged Keycloak instances

Skiepp:
It seems Realm creation with unmanaged keycloak instances works just fine once credential-<Keycloak_Name> secret is provided.

I suggest to change the error with a simple warning and let the operator proceed


Skiepp:
Lost interest

PR #467, opened 2022-01-27
Feature: Keycloak.externalDatabase.secretName

Nold360:
This Feature allows to specify the name of the secret holding the external database credentials. Making the operator compatible with e.g. postgres-operators.

## Checklist:
- [ ] Automated Tests - need help with that
- [ ] [Documentation](https://github.com/keycloak/keycloak-documentation) - will follow


vmuzikar:
@Nold360 Thank you for your contribution. The DB Secret is meant to be used just by the Keycloak Operator, not by any 3rd parties.

However, we are developing a new operator where we are taking a bit more dynamic approach with regards to the Secrets in config, see e.g. this [example](https://github.com/keycloak/keycloak/blob/19e19c9ddad64d23a29bb2eb3ec740187edc13c6/operator/src/main/resources/example-keycloak.yml#L10-L19). The first tech-preview will be [released in KC 18](https://www.keycloak.org/2022/03/releases).

PR #472, opened 2022-02-02
Reset admin password if credential secret recreated

jonathanvila:
## Related GH Issue
Closes KEYCLOAK-19841

## Checklist:
- [ ] Automated Tests


vmuzikar:
I believe this approach can't work. We can't rely on what's currently in the env vars.

A better solution would probably be at the KC side. We need to overwrite the existing admin creds. It's not ideal (what if somebody willingly changed the creds in KC) but it would work.

jonathanvila:
I agree.
The value in the Deployment/POD envs is coming from the secret, so if the
secret is deleted we lose those values.
To change the admin credentials from the operator using the API we need the
"old" admin credentials.
What do you suggest @vmuzikar

vmuzikar:
Closing this as we aren't going to implement this now (there's no simple solution for WF dist). The new operator has this resolved.

PR #361, opened 2021-07-12
support non-unique confidential clients

imykolenko:
## JIRA ID
https://issues.redhat.com/browse/KEYCLOAK-18721

## Additional Information
In this PR I change a name of k8s secret that holds client secret in order to support non-unique confidential clients with same clientId created in different realms.

This change is not backward compatible, hence ideas are welcome how/if this should be backward compatible.

PR #475, opened 2022-02-07
Fix client secret naming scheme to match docs

### LINKED ISSUE #473: Secret generated by explicitly defining a client secret in KeycloakClient CRD is not named as it is documented
### Describe the bug

In the documentation (https://www.keycloak.org/docs/latest/server_installation/index.html#_client-cr) it is stated that: "After a client is created, the Operator creates a Secret containing the Client ID and the client’s secret using the following naming pattern: `keycloak-client-secret-<custom resource name>`". However, the secret appears not to be named after CRD name itself but rather after the "clientId" field set for the client defined in the CRD. This causes problems when deploying clients with the same id in different realms, since Operator apparently tries to create one secret per each client but the secret names aren't unique due to the secret naming pattern (`keycloak-client-secret-<clientId>`).

### Version

16.1.0

### Expected behavior

Operator should name the created client secret with respect to pattern `keycloak-client-secret-<custom resource name>`

### Actual behavior

Operator names the created client secret with respect to pattern `keycloak-client-secret-<clientId>`

### How to Reproduce?

1. Deploy Keycloak Operator
2. Deploy a realm
3. Deploy a client to this realm with the following CRD 
 
`apiVersion: keycloak.org/v1alpha1
kind: KeycloakClient
metadata:
  name: someName
  labels:
    app: app=example-keycloak
spec:
  realmSelector:
     matchLabels:
      app: <matching labels for KeycloakRealm custom resource>
  client:
    clientId: someClientId
    secret: thisisasecret`

4. Check the secret to confirm it was named "keycloak-client-secret-someClientId" rather than "keycloak-client-secret-someName"

### Anything else?

_No response_

trevorloranger:
## Related GH Issue
Closes #473

## Additional Information

- Updated the naming scheme of the kubernetes secrets created by `KeylcoakClient` custom resources to now use the custom resource name (which matches Keycloak documentation) over the previously used `clientID`.
- `KeycloakClient` reconciliation process will create secrets using custom resource name, i.e.`keycloak-client-secret-<custom resource name>`, and also check for existence of secret using the old naming scheme, i.e. `keycloak-client-secret-<clientID>`, and delete appropriately. Put condition in place to avoid cyclical creation/deletion of secrets in case of `client_id == custom_resource_name`.
- Open to suggestions on approach and names used (deprecated was one that came to mind 🤷 )
- Tip of the hat to @imykolenko for their initial work in #361 👍 

## Verification Steps

See examples to reproduce issue caused by the `clientID` based naming scheme of the k8s secrets in [KEYCLOAK-18346](https://issues.redhat.com/browse/KEYCLOAK-18346).

## Checklist:
- [x] Automated Tests

### DIFF
--- a/pkg/controller/keycloakclient/keycloakclient_reconciler.go
+++ b/pkg/controller/keycloakclient/keycloakclient_reconciler.go
@@ -47,6 +47,12 @@ func (i *KeycloakClientReconciler) Reconcile(state *common.ClientState, cr *kc.K
 		desired.AddAction(i.getUpdatedClientSecretState(state, cr))
 	}
 
+	if state.DeprecatedClientSecret != nil {
+		// Delete client secret created using the previous naming scheme, i.e., keycloak-client-secret-<CLIENT_ID>.
+		// See GH issue #473 and KEYCLOAK-18346.
+		desired.AddAction(i.getDeletedDeprecatedClientSecretState(state, cr))
+	}
+
 	i.ReconcileRoles(state, cr, &desired)
 
 	i.ReconcileScopeMappings(state, cr, &desired)
@@ -292,10 +298,17 @@ func (i *KeycloakClientReconciler) getCreatedClientState(state *common.ClientSta
 	}
 }
 
+func (i *KeycloakClientReconciler) getDeletedDeprecatedClientSecretState(state *common.ClientState, cr *kc.KeycloakClient) common.ClusterAction {
+	return common.GenericDeleteAction{
+		Ref: state.DeprecatedClientSecret,
+		Msg: fmt.Sprintf("delete deprecated client secret %v/%v", cr.Namespace, cr.Spec.Client.ClientID),
+	}
+}
+
 func (i *KeycloakClientReconciler) getUpdatedClientSecretState(state *common.ClientState, cr *kc.KeycloakClient) common.ClusterAction {
 	return common.GenericUpdateAction{
 		Ref: model.ClientSecretReconciled(cr, state.ClientSecret),
-		Msg: fmt.Sprintf("update client secret %v/%v", cr.Namespace, cr.Spec.Client.ClientID),
+		Msg: fmt.Sprintf("update client secret %v/%v", cr.Namespace, cr.Name),
 	}
 }
 
@@ -310,7 +323,7 @@ func (i *KeycloakClientReconciler) getUpdatedClientState(state *common.ClientSta
 func (i *KeycloakClientReconciler) getCreatedClientSecretState(state *common.ClientState, cr *kc.KeycloakClient) common.ClusterAction {
 	return common.GenericCreateAction{
 		Ref: model.ClientSecret(cr),
-		Msg: fmt.Sprintf("create client secret %v/%v", cr.Namespace, cr.Spec.Client.ClientID),
+		Msg: fmt.Sprintf("create client secret %v/%v", cr.Namespace, cr.Name),
 	}
 }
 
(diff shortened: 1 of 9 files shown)


trevorloranger:
Hi @vmuzikar, wanted to tag you here since you initially commented on #361 and I included the clean-up of old secrets that you mentioned. Thanks!

pkg/controller/keycloakclient/keycloakclient_reconciler.go
```diff
@@ -47,6 +47,12 @@ func (i *KeycloakClientReconciler) Reconcile(state *common.ClientState, cr *kc.K
 		desired.AddAction(i.getUpdatedClientSecretState(state, cr))
 	}
 
+	if state.DeprecatedClientSecret != nil {
+		// Delete client secret created using the previous naming scheme, i.e., keycloak-client-secret-<CLIENT_ID>.
+		// See GH issue #473 and KEYCLOAK-18346.
+		desired.AddAction(i.getDeletedDeprecatedClientSecretState(state, cr))
```
andreaTP, line 53: Before adding this action we should be sure that the new ClientSecret have been already created, correct?
Probably something like `state.ClientSecret != nil`.

andreaTP (changes requested):
Thanks for the effort @trevorloranger trevorloranger
One note about the logic, and probably @vmuzikar will ask for an integration test of this functionality.

pkg/controller/keycloakclient/keycloakclient_reconciler.go
trevorloranger: Great point @andreaTP, but I think adding this will actually prevent old secret cleanup from occurring. As of my current understanding, which could be wrong:
1. The order these actions are queued implicitly addresses your scenario. If the "create new secret" action fails then none of the queued actions following it are executed, including the queued "delete old secret" action. Stated differently, the "delete old secret" action cannot occur in a single reconciliation loop unless it was [preceded](https://github.com/keycloak/keycloak-operator/blob/main/pkg/controller/keycloakclient/keycloakclient_controller.go#L160-L163) by a successful "create new secret" or "update new secret" action.
2. I believe adding this check will actually prevent the "delete old secret" from ever being queued. Specifically, during the reconciliation loop in which "create new secret" is performed, i.e. `state.ClientSecret == nil`, then unless the actual Kubernetes secret is updated, the reconciliation loop won't be triggered [again](https://github.com/keycloak/keycloak-operator/blob/main/pkg/controller/keycloakclient/keycloakclient_controller.go#L167) and subsequently, the "delete old secret" won't ever be scheduled.

Let me know your thoughts!

trevorloranger: @andreaTP let me know your thoughts ☝️

trevorloranger: Awesome, @andreaTP Is there anything else I need to do to get this merged? Thanks!

andreaTP: @trevorloranger for me is fine, you now need a review from @vmuzikar

trevorloranger:
@vmuzikar Hi there, do you know roughly when you'll be able to review this pull request? My team would love to get this fix in. Thank you! 🙏

vmuzikar:
@trevorloranger Hi, sorry, I've been really busy. Will try to look into this the next week. Thank you for your contribution!

trevorloranger:
@vmuzikar Hi, will you have a chance to look into this next week? Thanks again! 👍

PR #389, opened 2021-10-29
KEYCLOAK-17330 Update to latest stable keycloak-init container

sathieu:
The keycloak-init-container image is currently pinned to the "master"
shared tag, which has not been updated for two years.

This commit updates the image tag to use the latest released tag for the
container.

This container comes with a couple of benefits:

* Accurate URL parsing, the Python parsing has issues.
* Ability to use basic auth URLs for extensions.
* Much smaller image size.

Fixes #478

## Additional Information

- Reduce the vulnerabilities from 309 ( 101 High ) to none (https://quay.io/repository/keycloak/keycloak-init-container?tag=latest&tab=tags)
- Allows the use of private extensions via the use of a basic auth URL.

## Checklist:
- [ ] Automated Tests
- [x] [Documentation](https://github.com/keycloak/keycloak-documentation) changes if necessary
- [ ] (If required) Discussed on [Keycloak DEV](https://groups.google.com/forum/#!forum/keycloak-dev)

PR #484, opened 2022-02-13
Init container missing tag: latest to legacy

almahmoud:
## Related GH Issue
Related to https://github.com/keycloak/keycloak-operator/pull/389

## Additional Information
```
docker pull quay.io/keycloak/keycloak-init-container:latest
Error response from daemon: unknown: Tag latest was deleted or has expired. To pull, revive via time machine
```

Since I can't see tag `latest` but `legacy` is the most recent, assuming `legacy` was intended as the new default.


vmuzikar:
@almahmoud Thank you for your contribution. This is already addressed by #466.

PR #527, opened 2022-04-08
Operator creates ServiceMonitor even when metrics are not present

### LINKED ISSUE #521: Operator creates `ServiceMonitor` even when metrics are not present
### Describe the bug

When the operator [detects](https://github.com/keycloak/keycloak-operator/blob/77d126113104d45e29057f5c9e30c54f1a482763/pkg/controller/keycloak/keycloak_reconciler.go#L239-L243) that  `ServiceMonitor` object type (part of Prometheus operator) is present in the cluster, it creates a `ServiceMonitor` pointing to [`/auth/realms/master/metrics` endpoint](https://github.com/keycloak/keycloak-operator/blob/481ec6eaaa4d732243336faf006aeaa8d71571c6/pkg/model/service_monitor.go#L23) which comes from community [Keycloak Metrics SPI](https://github.com/aerogear/keycloak-metrics-spi).

When the extension is not explicitly installed, Prometheus complains about invalid targets.

Creation of `ServiceMonitor` can't be disabled.

### Version

17.0.1

### Expected behavior

We should introduce an optional switch to disable creation of `ServiceMonitor`.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

vmuzikar:
Closes #521


andreaTP (approved):
LGTM

PR #545, opened 2022-04-28
Service Account Roles Configuration

### LINKED ISSUE #544: Service Account Roles Configuration
### Description

Migrated from: https://issues.redhat.com/browse/KEYCLOAK-18294

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

vmuzikar:
Takes over #382

Closes #544

### DIFF
--- a/pkg/common/client_state.go
+++ b/pkg/common/client_state.go
@@ -11,24 +11,27 @@ import (
 )
 
 type ClientState struct {
-	Client                 *kc.KeycloakAPIClient
-	ClientSecret           *v1.Secret // keycloak-client-secret-<custom resource name>
-	Context                context.Context
-	Realm                  *kc.KeycloakRealm
-	Roles                  []kc.RoleRepresentation
-	DefaultRoleID          string
-	DefaultRoles           []kc.RoleRepresentation
-	ScopeMappings          *kc.MappingsRepresentation
-	AvailableClientScopes  []kc.KeycloakClientScope
-	DefaultClientScopes    []kc.KeycloakClientScope
-	OptionalClientScopes   []kc.KeycloakClientScope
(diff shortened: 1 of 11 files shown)


pkg/common/client_state.go
```diff
@@ -11,24 +11,27 @@ import (
...
+	Realm                   *kc.KeycloakRealm
+	Roles                   []kc.RoleRepresentation
+	DefaultRoleID           string
+	DefaultRoles            []kc.RoleRepresentation
+	ScopeMappings           *kc.MappingsRepresentation
+	AvailableClientScopes   []kc.KeycloakClientScope
+	DefaultClientScopes     []kc.KeycloakClientScope
+	OptionalClientScopes    []kc.KeycloakClientScope
+	DeprecatedClientSecret  *v1.Secret // keycloak-client-secret-<clientID>
+	Keycloak                kc.Keycloak
```
andreaTP, line 26: *if I followed the code correctly (TM 😅 ) you might not need `Keycloak` here.

You can simply pass `nil` to this function: https://github.com/keycloak/keycloak-operator/pull/545/files#diff-e2da6560e0d8eaff1fa4844b138e7e8193aa75cbd97b719d9430ac3c4f124628R105

and it "should" be not used anywhere else.

vmuzikar: Unfortunately, I'm afraid I can't use `nil` as `Keycloak` is used further here:
https://github.com/keycloak/keycloak-operator/blob/1d4eea5edb40835eb5284e0119fa68d024d1f61e/pkg/common/user_state.go#L120
and here:
https://github.com/keycloak/keycloak-operator/blob/86b110746968f5e3c47c998e2d9106cc5bf706a9/pkg/model/realm_credential_secret.go#L30

This is related to the user Secret which is not created for Service Accounts but its presence is still checked.

PR #567, opened 2022-07-11
Operator cannot reconcile StatefulSet if Selector is misconfigured

### LINKED ISSUE #566: Operator cannot reconcile StatefulSet if Selector is misconfigured
### Describe the bug

This issue is related to #512, see it also for steps to reproduce. However, the fix introduced in #513 is not complete as it requires the user to manually recreate the StatefulSet. 

### Version

18.0.0-18.0.2

### Expected behavior

Operator automatically recreates the StatefulSet to use correct `Selector` to avoid getting into deadlocked inconsistent state.

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

vmuzikar:
Close #566

### DIFF
--- a/pkg/controller/keycloak/keycloak_migrations.go
+++ b/pkg/controller/keycloak/keycloak_migrations.go
@@ -14,6 +14,7 @@ import (
 
 var errBackup = errors.New("migrate backup fails")
 var errNoMigrator = errors.New("migrator not found")
+var errSelectorCantBeMigrated = errors.New("statefulSet Selector mismatch; please use Recreate migration strategy")
 
 type Migrator interface {
 	Migrate(cr *v1alpha1.Keycloak, currentState *common.ClusterState, desiredState common.DesiredClusterState) (common.DesiredClusterState, error)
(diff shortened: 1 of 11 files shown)


vmuzikar:
@ASzc Please note that this requires Role changes to the OLM bundle as well.

pkg/controller/keycloak/keycloak_migrations.go
```diff
@@ -37,10 +38,45 @@ func GetMigrator(cr *v1alpha1.Keycloak) (Migrator, error) {
 }
 
 func (i *RecreateMigrator) Migrate(cr *v1alpha1.Keycloak, currentState *common.ClusterState, desiredState common.DesiredClusterState) (common.DesiredClusterState, error) {
-	if needsMigration(cr, currentState) {
+	deployment, deploymentIndex := findDeployment(&desiredState)
+
+	// We can't modify existing selector on StatefulSet.
+	// The selector might be wrongly set by e.g. RH-SSO 7.5.2.
```
andreaTP, line 44: @vmuzikar can you elaborate on when this situation might happen?

pkg/controller/keycloak/keycloak_migrations.go
```diff
@@ -37,10 +38,45 @@ func GetMigrator(cr *v1alpha1.Keycloak) (Migrator, error) {
...
+	// In such case, we need to recreate the StatefulSet
```
andreaTP, line 45: This looks to be a super dangerous operation, especially combined with OLM automatic updates: https://github.com/kubernetes/kubernetes/issues/50808#issuecomment-324746374

Would it be possible only to notify the user (e.g. using the status field) and leave the deletion to be a manual action performed by the user, so that it can be better scheduled/executed in production scenarios?

DGuhr (approved):
Approved, with the calculated risk we talked about in mind.

vmuzikar:
@andreaTP Thanks for the review! Based on our offline discussion yesterday, is it ok for you to approve?

andreaTP:
@vmuzikar , I miss seeing the agreed docs changes, do we have a separate ticket for it?

abstractj:
@vmuzikar would you like to follow up on @andreaTP comments?

vmuzikar:
@andreaTP Yes, we're discussing them offline for the moment. I can post a summary here after we're aligned.

PR #570, opened 2022-07-15
RH-SSO Deployment doesn't reflect serviceAccountName

### LINKED ISSUE #569: RH-SSO Deployment doesn't reflect `serviceAccountName`
### Describe the bug

Changes introduced in https://github.com/keycloak/keycloak-operator/pull/357 are not reflected in RH-SSO.

### Version

15.1.0-18.0.2

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

vmuzikar:
Closes #569

### DIFF
--- a/pkg/model/keycloak_deployment_test.go
+++ b/pkg/model/keycloak_deployment_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 type createDeploymentStatefulSet func(*v1alpha1.Keycloak, *v1.Secret, *v1.Secret) *v13.StatefulSet
-type reconcileRHSSODeployment func(*v1alpha1.Keycloak, *v13.StatefulSet, *v1.Secret, *v1.Secret) *v13.StatefulSet
+type reconciledDeployment func(*v1alpha1.Keycloak, *v13.StatefulSet, *v1.Secret, *v1.Secret) *v13.StatefulSet
 
 func TestKeycloakDeployment_testExperimentalEnvs(t *testing.T) {
 	testExperimentalEnvs(t, KeycloakDeployment)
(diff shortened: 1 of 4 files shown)


vmuzikar:
Also sneaked in a few fixes for `serviceAccoutName` as well.
1. `serviceAccoutName` is now reconciled, not just used at STS creation.
2. We don't assume the default `serviceAccoutName` at all. We just pass the empty string if not set in CR which is the default for the K8s client.

pkg/model/keycloak_deployment_test.go
```diff
@@ -489,20 +489,22 @@ func testServiceAccountSet(t *testing.T, deploymentFunction createDeploymentStat
...
-
-	//If serviceAccountName is not set in the cr, then the serviceAccountName should be default
+	statefulSet := deploymentFunction(cr, dbSecret, nil)
 
 	//when
-	serviceAccountName := deploymentFunction(cr, dbSecret, nil).Spec.Template.Spec.ServiceAccountName
 
-	assert.Equal(t, "default", serviceAccountName)
+	//If serviceAccountName is set in the cr, is should manifest itself in the statefulset
+	cr.Spec.KeycloakDeploymentSpec.Experimental.ServiceAccountName = "test2"
```
DGuhr, line 501: do we have e2e tests like for the new operator for this one? Would be interesting to know what happens when no ServiceAccountName is set (==nil or empty string i guess). I have too little context to evaluate the side effects, tbh, thats why I ask.

vmuzikar: In any case, the default (no `serviceAccountName` set) is actually tested by all E2E tests as we don't set it anywhere. :)

DGuhr: Thanks for the explanation. In that case I approve :)

pedroigor (approved):

